Issue #27, opened 2025-11-13
valid bootstrap issue：GET returned 429 Too Many Requests


sunxr:
INFO: Repository boringssl instantiated at:
  /Users/jamessun/Desktop/valdi/test_valdi/WORKSPACE:21:24: in <toplevel>
  /private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/valdi/bzl/workspace_prepare.bzl:5:23: in valdi_prepare_workspace
  /private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/valdi/bzl/dependencies.bzl:176:17: in setup_dependencies
Repository rule http_archive defined at:
  /private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/bazel_tools/tools/build_defs/repo/http.bzl:382:31: in <toplevel>
ERROR: An error occurred during the fetch of repository 'boringssl':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/bazel_tools/tools/build_defs/repo/http.bzl", line 131, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://boringssl.googlesource.com/boringssl/+archive/82f9853fc7d7360ae44f1e1357a6422c5244bbd8.tar.gz] to /private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/boringssl/temp12605984111393629039/82f9853fc7d7360ae44f1e1357a6422c5244bbd8.tar.gz: GET returned 429 Too Many Requests
ERROR: no such package '@@boringssl//': java.io.IOException: Error downloading [https://boringssl.googlesource.com/boringssl/+archive/82f9853fc7d7360ae44f1e1357a6422c5244bbd8.tar.gz] to /private/var/tmp/_bazel_jamessun/7c6d97620daed444968d7121f5353ee3/external/boringssl/temp12605984111393629039/82f9853fc7d7360ae44f1e1357a6422c5244bbd8.tar.gz: GET returned 429 Too Many Requests
[...]

PR #31, opened 2025-11-13
fix: add github mirror url for boringssl to handle rate limiting

c4nzin:
issue : https://github.com/Snapchat/Valdi/issues/27

## Problem
Boringssl downloads fail with 429 Too Many Requests from Google's server

## Solution
Add GitHub as fallback URL in http_archive

## Changes
- Updated bzl/dependencies.bzl to use urls array with GitHub mirror


clholgat:
Thanks for the proposed fix, this is now obsolete with bzlmod.

PR #32, opened 2025-11-14
Add Fedora setup option depending of which distro you are in

diegosoriarios:
Trying to install Valdi on my machine I got some problems trying to build, looking at the code I realized it was because I am using Fedora and the config was built to work with debian based distros.
This pull request add Linux distribution support in the CLI setup process, providing tailored installation commands and setup logic for both Ubuntu and Fedora systems. The changes improve the user experience by detecting the user's Linux distribution and running the appropriate dependency installation steps, rather than assuming all users are on Ubuntu. Additionally, the doctor command now gives more accurate instructions for resolving missing dependencies based on the detected distribution.

**Linux distribution detection and setup:**

* Added detection for Ubuntu and Fedora distributions in `linuxSetup.ts`, and refactored the main setup logic to delegate to new, distro-specific setup functions (`ubuntuSetup` and `fedoraSetup`). If an unsupported distro is detected, a clear error message is shown.
* Implemented `ubuntuSetup.ts` and `fedoraSetup.ts` to encapsulate the installation steps for each distribution, including dependency installation and Java runtime checks.

**Improved dependency installation instructions:**

* Updated the `ValdiDoctor` class to provide more accurate fix commands for missing dependencies, using `sudo apt-get` for Ubuntu, `sudo dnf` for Fedora, and a generic message for other distributions. 
* Imported the new distro detection helpers (`isFedora`, `isUbuntu`) into the doctor command for use in generating these instructions.

If anyone have any other changes feel free to change.

### DIFF
--- a/npm_modules/cli/src/commands/doctor.ts
+++ b/npm_modules/cli/src/commands/doctor.ts
@@ -45,6 +45,7 @@ import { BazelClient } from '../utils/BazelClient';
 import { checkCommandExists, runCliCommand } from '../utils/cliUtils';
 import { makeCommandHandler } from '../utils/errorUtils';
 import { wrapInColor } from '../utils/logUtils';
+import { isFedora, isUbuntu } from '../setup/linuxSetup';
 
 /** Discord support link for troubleshooting */
 const DISCORD_SUPPORT_URL = 'https://discord.gg/uJyNEeYX2U';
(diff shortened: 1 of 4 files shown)


clholgat:
Thanks for the proposed fix. We've since expanded Linux support.

PR #35, opened 2025-11-14
fix: don't only rely on apt if on Linux

ThatOneCalculator:
Right now, valdi's setup commands assume all Linux setups have `apt`/`apt-get`. While this isn't a complete implementation for every distro, this should solve the problem of trying to install all dependencies with `apt` even if it's not present on the system.


ThatOneCalculator:
Supersceded by #41

PR #37, opened 2025-11-14
fix: disable bazel sandboxing for Linux

ThatOneCalculator:
- Parity with macOS (not sure why this was different in the first place)
- Fixes `valdi bootstrap` on Arch Linux

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -22,9 +22,8 @@ common --define enable_web=true
 build --tool_java_language_version=11
 build --java_language_version=11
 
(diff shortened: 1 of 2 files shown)


clholgat:
This is a little heavy handed. If you're still having issues on Arch Linux, can you file a ticket with more information? Let's see if there's another solution.

ThatOneCalculator:
> This is a little heavy handed. 

Is it really? macOS disables sandboxing too, and `/var/tmp` isn't a valid temp directory for tmpfs distros

PR #43, opened 2025-11-15
Add macOS Intel (x64) Support for Valdi Hotreload

japheth-waswa:
### Summary

This PR adds full hotreload support for macOS Intel (x86_64) machines by enabling the Valdi compiler companion to load to load a native SQLite binding for intel platforms

### 🧠 Motivation

`Valdi hotreload` previously failed on macOS Intel systems with:

```bash
Unsupported platform and arch combination (darwin on arch x64)
```

This was caused by:

- Missing native SQLite binding for macOS x86_64
- Platform detection logic supporting only arm64 and Linux x64

This PR enables Intel developers to use Valdi without requiring ARM hardware.

### 🆕 Changes

- Added platform support for "darwin" && "x64" in the file `compiler/companion/src/cache/SQLiteCompilationCache.ts`
```ts
else if (platform === 'darwin' && arch === 'x64') {
    nativeBinding = require('../sqlite_bindings/better_sqlite3_macos_x64.node');
  }
```
- Added better_sqlite3_macos_x64.node native binding
Compiled specifically for macOS Intel using Node v22.21.1 to match Valdi's embedded Node runtime.

###  🔧 Build Environment Used for this PR

- Node.js Version: v22.21.1
- Node ABI Version: 127
- Architecture: x86_64
- OS: macOS 15.1 (Sequoia) Intel

### 🔧 Build Procedure

```bash
nvm install 22.21.1
nvm use 22.21.1
git clone https://github.com/WiseLibs/better-sqlite3
cd better-sqlite3
npm install
npm run build-release
```

Produced file

```
build/Release/better_sqlite3.node
 → renamed to better_sqlite3_macos_x64.node
 → placed inside compiler/companion/src/sqlite_bindings/
```

### 🧪 Testing

✔ valdi hotreload works on macOS Intel
✔ Builds correctly using Bazel
✔ Local projects bootstrap and sync as expected
✔ SQLiteCompilationCache initializes successfully

### 📌 Compatibility

This change is:
- fully backward-compatible
- safe for macOS ARM
- safe for Linux
- no impact on iOS/Android build pipelines
- introduces no behavioral changes outside Intel runtime

###  🙌 Notes

If Valdi upgrades its embedded Node version in the future, maintainers need to rebuild both:
- better_sqlite3_macos_x64.node
- better_sqlite3_macos_arm64.node

using the same embedded Node version & ABI.


clholgat:
Thanks for the proposed fix but we aren't planning to support intel Macs.

PR #47, opened 2025-11-17
Add README Chinese Suppoort

Jarvis636431:


clholgat:
Thanks for this translation but the README has changed significantly and we aren't able to verify this translation.

Issue #1, opened 2025-11-07
Issue running setup CLI


nandorojo:
Hey! Thanks for open sourcing this, excited to play with it. I just got this error running the setup script:

```sh
cd Valdi/npm_modules/cli/

npm run cli:install

valdi dev_setup
```

The `valdi dev_setup` command gave me this error:

```sh
openjdk@11
For the system Java wrappers to find this JDK, symlink it with
  sudo ln -sfn /opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk-11.jdk

openjdk@11 is keg-only, which means it was not symlinked into /opt/homebrew,
because this is an alternate version of another formula.

If you need to have openjdk@11 first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/openjdk@11/bin:$PATH"' >> ~/.zshrc

For compilers to find openjdk@11 you may need to set:
  export CPPFLAGS="-I/opt/homebrew/opt/openjdk@11/include"
==> git-lfs
Update your git config to finish installation:

  # Update global git config
  $ git lfs install

  # Update system git config
  $ git lfs install --system
Setting up JDK...
- sudo ln -sfn /opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk-11.jdk
✔ Updating /Users/.../.zshrc to include PATH, JAVA_HOME...
Setting up Android SDK...
Installing Android platform...
- /Users/.../Library/Android/sdk/cmdline-tools/latest/bin/sdkmanager --install 'platforms;android-35'
zsh:2: no such file or directory: /Users/.../Library/Android/sdk/cmdline-tools/latest/bin/sdkmanager
Error
    at ChildProcess.<anonymous> (/Users/.../Developer/Valdi/npm_modules/cli/dist/utils/cliUtils.js:49:24)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

Any ideas here? Maybe there's some other Android prerequisite? Thanks!

busybox11:
Have you followed the [prerequisites setup guide](https://github.com/Snapchat/Valdi/blob/main/docs/setup/macos_setup.md) and sourced your rc file / restarted your shell?

[...]

PR #3, opened 2025-11-08
Setup dev env: Android CLI tools docs and Valdi CLI improvements

busybox11:
Since devs will need to setup some parts of the Android SDK themselves beforehand, as instructed in the MacOS and Linux setup guide, this PR adds a notice to install the latest Android SDK CLI Tools from the SDK Manager as well.

This PR also removes the manual download, unzip, copy and handling of the `sdkmanager`, but rather assumes the user has followed the setup guide and installed the tools from SDK Manager.

Fixes #1

PR #52, opened 2025-11-18
Update macos_setup.md

ecoreng:
Addresses the issue in https://github.com/Snapchat/Valdi/issues/1


ecoreng:
Dup of https://github.com/Snapchat/Valdi/pull/3

PR #78, opened 2025-12-03
lib/utils/crypto: use 32 byte key size instead of 16

iurjscsi1101500:
This is done to improve security.


clholgat:
Closing: this change modifies crypto key sizes without context or justification, and the PR has conflicts. The existing key sizes are intentional.

PR #89, opened 2026-02-21
fix: skip runtime IR emission for const enum declarations

AugustHagedal:
## Description

Found a bug in `processEnumDeclaration()` — it was emitting a runtime object and property assignments for `const enum` declarations, which it shouldn't. Const enums are compile-time only; TypeScript inlines their values at every usage site via `emitResolver.getConstantValue()`, so there's nothing to emit at the declaration site.

The fix is a small early return when the `Const` modifier is detected. The TODO comment that was sitting there is gone too.

## Type of Change
- [x] Test improvement

## Testing
- [x] Tests pass locally

### Testing Details

Added two tests to `NativeCompiler.spec.ts` alongside the existing const enum test:

- Inline integer const enum: checks `Color.Green` becomes `storeint 1` with no `newobject` in the output
- Inline string const enum: checks `Direction.Up` becomes `storestring 'UP'` with no `newobject`

All 118 tests pass.

## Checklist
- [x] No breaking changes


clholgat:
This looks good, we're going to run this through some additional testing.

clholgat (approved):

PR #90, opened 2026-02-21
fix: escape single quotes and backslashes in IR string output

AugustHagedal:
## Summary

- Strings containing `'` or `\` in their value broke the IR parser because `appendQuotedString` emitted them unescaped inside single-quote delimiters (a TODO comment had flagged this since the file was written)
- The `LiteralString` IR case had its own inline template that bypassed `appendQuotedString` entirely, so it also emitted unescaped strings

## Changes

- Add `escapeIRString()` helper in `IRtoString.ts` that escapes `\`, `'`, `\n`, `\r`, and `\t`
- Apply it in `appendQuotedString()`, removing the long-standing `TODO(simon): Escape quotes` comment
- Fix the `LiteralString` case to route through `appendQuotedString` instead of its own inline template literal
- Add two regression tests: single quote and backslash in string literals (120 tests, all passing)

## Test plan

- [x] Run existing test suite — all 118 tests continue to pass
- [x] New test: `escapes single quotes in string literals` — verifies `storestring 'it\'s a test'`
- [x] New test: `escapes backslashes in string literals` — verifies `storestring 'C:\\Users\\file'`

```
Tests: 120 passed, 120 total
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


clholgat:
This looks good, running this through some extra testing.

PR #123, opened 2026-07-23
Add linear and radial gradients for shape fills

scorsin-oai:
## Description
This change adds support for fillGradient on the shape element.

## Type of Change
- [ ] Test improvement
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/snap_drawing/src/snap_drawing/cpp/Layers/ShapeLayer.cpp
+++ b/snap_drawing/src/snap_drawing/cpp/Layers/ShapeLayer.cpp
@@ -28,7 +28,13 @@ void ShapeLayer::onInitialize() {
 
 void ShapeLayer::onDraw(DrawingContext& drawingContext) {
     const auto& path = getActivePath();
-    drawingContext.drawPaint(_fillPaint, path);
+    auto fillPaint = _fillPaint;
+    if (_fillGradientWrapper.hasGradient()) {
+        fillPaint.setColor(Color::black());
+        _fillGradientWrapper.update(path.getBounds().value_or(drawingContext.drawBounds()));
+        _fillGradientWrapper.applyToPaint(fillPaint);
+    }
+    drawingContext.drawPaint(fillPaint, path);
     drawingContext.drawPaint(_strokePaint, path);
 }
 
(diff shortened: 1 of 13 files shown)


scorsin-oai:
Will add iOS impl before merging the PR

valdi/src/java/com/snap/valdi/attributes/impl/ShapeViewAttributesBinder.kt
```diff
@@ -52,6 +53,23 @@ class ShapeViewAttributesBinder : AttributesBinder<ShapeView> {
...
     }
 
+    private fun applyFillGradient(view: ShapeView, value: Array<Any>, animator: ValdiAnimator?) {
+        val valdiGradient = ValdiGradient.fromGradientData(value)
+
+        if (valdiGradient.colors.isEmpty()) {
+            view.resetFillGradient()
+        } else if (valdiGradient.colors.size == 1) {
+            view.resetFillGradient()
+            view.setFillColor(valdiGradient.colors[0])
```
afan-sc: Seems like the one-color path replaces the shape's stored `fillColor`. Android calls `setFillColor()` here, while iOS similarly assigns the gradient color to `_fillColor`. When the gradient is removed, the reset path restores the gradient color rather than the original bound fill color.

tested this by starting with `fillColor="green"`, applying a one-color red gradient, and then removing the gradient. The shape stayed red on both Android and iOS. Just wanted to double check if that was intended

valdi/src/java/com/snap/valdi/views/ShapeView.kt
```diff
@@ -122,11 +154,114 @@ class ShapeView(context: Context) : View(context), ValdiRecyclableView {
...
+                    y0,
+                    x1,
+                    y1,
+                    gradient.colors,
+                    gradient.locations,
+                    Shader.TileMode.CLAMP,
+                )
+            }
+            GradientDrawable.RADIAL_GRADIENT -> {
+                fillPaint.shader = RadialGradient(
```
afan-sc: When either shape dimension is zero, the calculated radius is also zero. Android's `RadialGradient` requires a radius greater than zero, so constructing the shader in that state crashes.

I double-checked this by changing the shape's width and height to zero independently. Android crashed in both cases, but iOS handled them.

Since zero bounds can occur temporarily during layout, could we skip creating the shader until valid bounds return?

valdi/src/java/com/snap/valdi/views/ShapeView.kt
```diff
@@ -122,11 +154,114 @@ class ShapeView(context: Context) : View(context), ValdiRecyclableView {
...
+                    fillGradientBounds.centerX(),
+                    fillGradientBounds.centerY(),
+                    min(fillGradientBounds.width(), fillGradientBounds.height()) / 2.0f,
```
afan-sc: It looks like Android calculates one circular radius using `min(width, height) / 2`, while iOS and SnapDrawing scale the radial gradient with the non-square bounds. This seems to produce different results for rectangular shapes. Tested the same red-to-blue gradient in a `2:1` shape. Android rendered a circle, while iOS and SnapDrawing rendered an ellipse.

Is the Android-specific behavior intentional? If not, could we align it with the other backends and add a non-square radial test?

snap_drawing/src/snap_drawing/cpp/Layers/ShapeLayer.cpp
```diff
@@ -28,7 +28,13 @@ void ShapeLayer::onInitialize() {
 
 void ShapeLayer::onDraw(DrawingContext& drawingContext) {
     const auto& path = getActivePath();
-    drawingContext.drawPaint(_fillPaint, path);
+    auto fillPaint = _fillPaint;
+    if (_fillGradientWrapper.hasGradient()) {
+        fillPaint.setColor(Color::black());
+        _fillGradientWrapper.update(path.getBounds().value_or(drawingContext.drawBounds()));
```
afan-sc: It looks like SnapDrawing sizes the gradient using the path bounds, while Android and iOS use the full shape layer. This can produce different results when the path does not occupy the complete layer.

I double-checked with a `120×100` layer whose path only occupies the top half. The Android and iOS versions ended around purple, while SnapDrawing reached blue.

Is the path-relative behavior expected here? If not, would using the drawing/layer bounds make the backends more consistent?

scorsin-oai:
@afan-sc fixed!

valdi/src/java/com/snap/valdi/views/ShapeView.kt
```diff
@@ -122,11 +154,130 @@ class ShapeView(context: Context) : View(context), ValdiRecyclableView {
...
     }
 
+    private fun updateFillGradientShader() {
+        val gradient = fillGradient ?: return
+        if (!fillGradientDirty && fillGradientWidth == width && fillGradientHeight == height) {
+            return
+        }
+
+        if (width <= 0 || height <= 0) {
+            fillPaint.shader = null
```
afan-sc, line 183: I think clearing the shader here leaves the cached dimensions unchanged. When the shape returns to its previous size, the cache check above can skip rebuilding the shader, causing the shape to render with the black fallback color instead of its gradient.

I double-checked this in Composer Playground on Android by changing a radial-gradient shape from its valid size to zero width or height and then restoring it. It did not crash, but the restored shape turned black. I also confirmed the same valid → zero → valid sequence with a focused regression test.

Would it make sense to mark `fillGradientDirty = true` when clearing the shader and add the recovery sequence to `radialFillGradientWaitsForNonzeroBounds()`?

scorsin-oai: Fixed in 156b793e. Clearing the shader for zero-sized bounds now also marks `fillGradientDirty = true`, so restoring the previous dimensions rebuilds the gradient instead of falling back to black. I also extended `radialFillGradientWaitsForNonzeroBounds()` to cover both valid → zero width → valid and valid → zero height → valid; the regression reproduces before the fix and passes afterward.

PR #125, opened 2026-07-23
Add renderer memoization and reconciliation fixes

scorsin-oai:
## Description
This change adds the `remember()` API which works kind of like React's `useMemo()` or Jetpack Compose `remember()`. Unlike React, this API doesn't just work on the nearest ancestor component, it works on the nearest node that is at the top,.

It also fixes a reconcilation bug which could cause duplicate keys when it shouldn't, and it fixes a crash that could occur when a node has its viewClass being changed.

## Type of Change
- [ ] Test improvement
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/src/valdi_modules/src/valdi/valdi_core/src/Renderer.ts
+++ b/src/valdi_modules/src/valdi/valdi_core/src/Renderer.ts
@@ -33,6 +33,16 @@ import { trace } from './utils/Trace';
 const EMPTY_OBJECT = Object.freeze({});
 const EMPTY_ARRAY = Object.freeze([]) as [];
 
+interface RememberSlot<T = unknown> {
+  value: T;
+  keys: unknown[];
+}
+
+interface RememberState {
+  slots: RememberSlot[];
+  slotIndex: number;
+}
+
 interface NodeChildren<T> {
   childByKey: { [key: string]: T };
   children: T[];
@@ -67,6 +77,8 @@ interface VirtualNode extends Node {
 
   // Bridge instance, will be set if the virtual node was requested externally.
   bridge?: VirtualNodeBridge;
+
+  rememberState?: RememberState;
 }
 
 interface RenderedElement {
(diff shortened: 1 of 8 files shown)


beaucollins (approved):
Importing PR for internal review.

src/valdi_modules/src/valdi/valdi_core/src/Renderer.ts
```diff
@@ -1611,6 +1647,36 @@ export class Renderer implements IRenderer {
...
+
+    const slotIndex = rememberState.slotIndex++;
+    const slots = rememberState.slots;
+    const slot = slots[slotIndex] as RememberSlot<T> | undefined;
+    if (slot && rememberKeysEqual(slot.keys, keys)) {
+      return slot.value;
+    }
+
+    const value = factory();
+    slots[slotIndex] = {
```
beaucollins, line 1673: `rememberState.slots` is only ever appended to or overwritten in place — nothing shrinks it. `pushVirtualNode` resets `slotIndex` to 0 (line 1081) and the array is only released when the node is destroyed (`destroyVirtualNode`, line 1493). So when a render executes fewer `remember` calls than a previous render did — a conditional `remember`, or a loop whose iteration count drops — the trailing slots keep their `value` and `keys` alive, along with whatever those closures capture, for the remaining lifetime of the virtual node.

Retention is bounded by the node's peak slot count rather than unbounded, but for a long-lived node whose `remember` count fluctuates it is dead memory held indefinitely.

Truncating in `popVirtualNode` fixes it:

```ts
private popVirtualNode() {
  this.nodeStack.pop();
  const newCurrentNode = this.nodeStack[this.nodeStack.length - 1];

  const node = this.currentNode!;
  const rememberState = node.rememberState;
  if (rememberState && rememberState.slots.length > rememberState.slotIndex) {
    rememberState.slots.length = rememberState.slotIndex;
  }

  this.currentNode = newCurrentNode;
  // ...
```

One caveat worth deciding explicitly: slots are matched positionally, so a conditional `remember` also misaligns slots between renders and can hand back a value memoized for a different call site. That is a correctness issue independent of the retention above. If the intended contract is React's rules-of-hooks (`remember` must be called unconditionally, in the same order, every render), it is worth documenting on the public API in `Remember.ts` — and in that case truncation only matters for the shrinking-loop case rather than the conditional one.

scorsin-oai: I am aware of this limitation but I specifically chose not to fix it to keep the code simple given that this is a corner case, will be inapplicable most of the time, and the fix requires additional book keeping / code to run on every node being popped regardless if they use remember() or not. 

Yes, remember calls must be called in the same order on every render.

PR #127, opened 2026-07-24
Add MessageChannel and aggressive worker termination

scorsin-oai:
## Description
This change adds the MessageChannel and MessagePort API, modeled against how Web works.
It also makes `worker.terminate()` more aggressively terminate the worker. This allows worker code that runs in an infinite loop to be terminated. Note that in the JSCore impl it's not perfect because JSCore does not expose public APIs that allows us to implement this safely, so we have to rely on waiting until the JS code goes into our code during a native call.

## Type of Change
- [ ] Test improvement
- [ ] Other (please describe)

## Testing

## Checklist


valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    void start(const JavaScriptMessagePort& handle);
+    void close(const JavaScriptMessagePort& handle);
+    void enqueue(const Ref<JavaScriptMessage>& message);
+
+private:
+    mutable Mutex _mutex;
+    Weak<JavaScriptMessagePortEndpoint> _peer;
+    Weak<JavaScriptRuntime> _ownerRuntime;
+    Weak<JavaScriptMessagePort> _handle;
```
clholgat, line 67: `_handle` being weak means nothing roots an active port. `newObjectFromNativeClass` produces an ordinary collectible JS object (on JSCore it's a plain `JSObjectMake` + private data, no registry), so a port whose only liveness is a registered `onmessage` can be collected. `~JavaScriptMessagePort()` -> `close()` then sets `_closed`, clears `_messages` and resets `_ownerRuntime`, after which the peer's `enqueue()` early-returns and `postMessageFromJavaScript()` no-ops on the null peer runtime. The messages disappear with no error.

The HTML spec requires the opposite: a port that is entangled, enabled, and has a message event listener must not be collected.

```ts
function connect(): MessagePort {
  const ch = new MessageChannel();
  ch.port2.onmessage = e => handle(e.data);
  return ch.port1;            // ch and ch.port2 are unreachable from here
}
const p = connect();
p.postMessage('hi');          // silently dropped once port2 is collected
```

Every spec in `MessageChannel.spec.ts` keeps the channel in scope for the duration of the test, so the suite can't catch this. Suggest the endpoint hold a strong ref to the handle while it is started and has a handler, released on `close()` / `transfer()`.

This is the one item I'd want fixed before merge -- it's silent data loss in the lifetime model of the new API.

valdi/src/valdi/jscore/JavaScriptCoreContext.cpp
```diff
@@ -1330,22 +1331,33 @@ void JavaScriptCoreContext::willEnterVM() {
     _enterVmCount++;
 }
 
+void JavaScriptCoreContext::requestExecutionTermination() {
+    IJavaScriptContext::requestExecutionTermination();
```
clholgat, line 1335: A no-op override makes the per-engine asymmetry easy to miss. On JSCore there's no interrupt hook, so `requestExecutionTermination()` only takes effect at the native-call checkpoints in `JSCoreCustomClasses.cpp` -- a worker spinning in a pure-JS loop that never calls into native can't be terminated here.

You call this out in the description and I agree there's no clean JSCore API for it. What I want to flag is the consequence across engines: QuickJS gets real preemption through the interrupt handler, Hermes gets it through `triggerTimeoutAsyncBreak` + the new `emitAsyncBreakCheck`, and JSCore gets neither -- while `getAuto()` resolves to JSCore on iOS. So the engine paying the `emitAsyncBreakCheck` codegen cost is not the one where termination is weakest.

Not asking you to solve the JSCore side. Could you document the guarantee per engine (here or in the Worker docs) so callers don't read `terminate()` as a hard kill?

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
```diff
@@ -0,0 +1,511 @@
...
+    auto data = valueToJSValue(entry.jsContext,
+                               message.getData(),
+                               &nativeObjectResolver,
+                               ReferenceInfoBuilder().withProperty(STRING_LITERAL("data")),
+                               entry.exceptionTracker);
+    if (!entry.exceptionTracker) {
+        return entry.jsContext.newUndefined();
+    }
+
+    auto event = entry.jsContext.newObject(entry.exceptionTracker);
```
clholgat, line 514: Missing an `if (!entry.exceptionTracker)` check after `newObject` -- every other JS call in this function checks. If `newObject` fails, `event.get()` is still handed to both `setObjectProperty` calls below, and the two `setObjectProperty` calls are themselves unchecked before `return event`.

`callHandler` does check the tracker before invoking the handler, so the blast radius is small, but operating on a failed value is engine-dependent behaviour and it's inconsistent with the convention the rest of the file follows.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
```diff
@@ -0,0 +1,511 @@
...
+        }
+    }
+
+    std::vector<Ref<ValdiObject>> transferredPortSources;
+    transferredPortSources.reserve(ports.size());
+    std::vector<Ref<JavaScriptMessagePortEndpoint>> transferredPorts;
+    transferredPorts.reserve(ports.size());
+    for (const auto& port : ports) {
+        transferredPortSources.emplace_back(port);
+        transferredPorts.emplace_back(port->transfer(targetRuntime));
```
clholgat, line 341: Validation and mutation are two separate passes, each taking the lock per port, so between `validateTransfer()` above and `transfer()` here the port's state can in principle change -- and `Endpoint::transfer` guards that with `SC_ASSERT(!_closed && currentHandle != nullptr && currentHandle.get() == &handle)`, which compiles out in release.

I believe this is safe today: both passes run on the owning runtime's JS thread with no yield between them, so nothing can interleave. The concern is that the invariant isn't stated anywhere, and if it's ever broken the release-build failure mode is silent state corruption rather than a crash.

Could you either document the thread-affinity requirement on `make()` / `transfer()`, or promote the assert to a checked `Result` so a violation surfaces?

src/valdi_modules/src/valdi/worker/src/Worker.ts
```diff
@@ -13,15 +13,15 @@ export class Worker {
     this.nativeWorker = runtime.createWorker(url);
   }
 
-  public set onmessage(func: OnMessageFunc<unknown>) {
+  public set onmessage(func: (event: MessageEvent<unknown>) => void) {
```
clholgat: This types the handler as the DOM `MessageEvent`, but what the runtime delivers is a plain object carrying only `data` and `ports` (`makeJavaScriptMessageEvent` in `JavaScriptMessagePort.cpp`). Since the Valdi tsconfig bases set `"lib": ["dom", ...]`, that means `event.origin` / `event.lastEventId` / `event.source` all typecheck and evaluate to `undefined`, `event instanceof MessageEvent` is false, and `event.preventDefault()` typechecks and throws.

Same applies to consumers of `NativeMessageEvent`. Either narrow the public type to the fields actually delivered, or populate the rest of the shape -- the narrow type seems preferable to growing the event object.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
```diff
@@ -0,0 +1,511 @@
...
+            // Transferring the destination port makes the message undeliverable.
+            return;
+        }
+    }
+
+    std::unique_lock<Mutex> lock(_mutex);
+    if (_closed) {
+        return;
+    }
+    _messages.emplace_back(message);
```
clholgat, line 152: A port that's transferred but never materialised on the receiving side leaves this queue unbounded: `_handle` stays empty, `scheduleNextMessage()` early-returns every time, and the peer keeps appending. That's reachable whenever the message carrying the port is never delivered -- the receiver never registers a handler, or drops the event.

Web has the same shape so this is arguably conformant, and I don't think it blocks anything. But it's an unbounded native allocation with no diagnostic in a long-lived app. Worth a soft cap plus a `VALDI_WARN` so it's discoverable?

clholgat (commented):
Went through the `MessagePort` / `MessageChannel` implementation in detail. The core is careful work -- generation counters to invalidate scheduled dispatches across transfers, all-before-any transfer-list validation, port identity preserved when a port appears both in the transfer list and embedded in the message data, FIFO ordering, implicit start on `onmessage` assignment, idempotent `close()`. The tests cover each of those explicitly, which made the review much easier.

One issue I'd want resolved before this merges -- the port lifetime / GC comment on `JavaScriptMessagePort.hpp`. Four smaller ones inline; those are comments rather than blockers.

Separately, and not something you need to act on: adding `MessageChannel` as a global flips which code path `jasmine.js` uses for `clearStack()` (it prefers `setImmediate`, then `MessageChannel`, then `setTimeout`, and Valdi defines no `setImmediate`). Combined with the `willExitVM` reordering, that changes microtask timing for existing suites, so I'll run this against the full downstream test suite on my side before importing.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
scorsin-oai: Documented the invariant on both JavaScriptMessage::make and JavaScriptMessagePortEndpoint::transfer: validation and mutation must happen without yielding during the same source-runtime JavaScript-thread turn. Existing call paths satisfy that invariant.

valdi/src/valdi/jscore/JavaScriptCoreContext.cpp
scorsin-oai: Documented directly on Worker.terminate(): Hermes and QuickJS can interrupt running JavaScript, whereas JavaScriptCore only observes termination at native-call boundaries and cannot interrupt a pure JavaScript loop.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
scorsin-oai: Agreed that this can grow without bound when a transferred destination never materializes. I left queue behavior unchanged because it matches browser MessagePort semantics, and a cap or warning threshold would introduce a separate delivery/diagnostic policy. We can revisit observability independently if this becomes operationally relevant.

src/valdi_modules/src/valdi/worker/src/Worker.ts
scorsin-oai: Fixed: Worker.onmessage and the worker/channel tests now use the existing NativeMessageEvent and NativeMessagePort interfaces, so callbacks expose only the data and ports fields actually supplied by native code.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.hpp
scorsin-oai: Fixed in ad0c194b. Listener-backed ports are now retained through a context-attached strong reference, which is released when the listener is removed, the port is closed or transferred, or its context is disposed. A forced-GC integration test reproduced the previous failure on QuickJS, QuickJS+TSN, and JavaScriptCore, and now passes across all four engine configurations.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
scorsin-oai: Fixed: makeJavaScriptMessageEvent now checks exceptionTracker immediately after newObject and after each data/ports setObjectProperty call, returning undefined on failure.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.hpp
clholgat: Verified the rooting and teardown path end to end. Retention is released on close(), transfer(), handler removal, and context disposal; because the retained ref keeps the handle alive, `~JavaScriptMessagePort` can only run once retention is already gone, so the close() in the destructor is a no-op on that path (no double-free); and the `strongSmallRef(this)` self-pin in dispose() correctly covers the inner-handle release cascading back and freeing the CAMP mid-dispose. The forced-GC integration test hits the exact previously-broken path. Clearing this as the blocker — thanks for the careful fix.

valdi/src/valdi/jscore/JavaScriptCoreContext.cpp
clholgat: The per-engine note you added on `terminate()` in Worker.ts is exactly what I was after — callers won't read it as a hard kill on JSCore now. Resolved.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
clholgat: Confirmed — checks the tracker after `newObject` and after each `setObjectProperty`, returning undefined on failure. Matches the convention the rest of the file follows.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
clholgat: Documenting the no-yield-within-the-turn invariant on make()/transfer() works for me — the existing call paths satisfy it, and the failure mode is now called out for anyone adding a new one. Thanks.

src/valdi_modules/src/valdi/worker/src/Worker.ts
clholgat: Confirmed — handlers are now typed against `NativeMessageEvent`/`NativeMessagePort`, so the phantom DOM fields (`origin`/`lastEventId`/`source`) and `preventDefault()` are gone from the surface.

valdi/src/valdi/runtime/JavaScript/JavaScriptMessagePort.cpp
clholgat: Agreed — leave it. Browser MessagePort has the same shape, and a cap/warning is a separate observability policy that shouldn't ride in on this PR. Not a blocker.

clholgat (commented):
Re-reviewed `ad0c194b`. The lifetime fix is the right shape — an active, listener-backed port now roots itself through a context-attached strong reference (`endpoint._retainedHandle` → `ContextAttachedMessagePort` → handle → `endpoint`), and that cycle is broken on `close()`, `transfer()`, handler removal, and context disposal, which matches the HTML requirement that an entangled, enabled port with a listener stays alive.

I traced the teardown:
- Because retention keeps the handle alive, `~JavaScriptMessagePort` can only run once retention is already released, so the `close()` in the destructor is a no-op on that path — no double-free.
- The `strongSmallRef(this)` self-pin in `ContextAttachedMessagePort::dispose()` is necessary and correct: releasing the inner handle can cascade back through `~handle → close() → endpoint drops _retainedHandle` and free the CAMP mid-dispose; the pin covers that window.
- The forced-GC integration test in `Runtime_tests.cpp` exercises the exact previously-broken path (GC while listening still delivers; close + GC collects), across engine configs.

H1 is resolved — clearing my blocker. M1 (tracker checks), M2 (documented thread-affinity invariant), M3 (`NativeMessageEvent`/`NativeMessagePort` types), and the per-engine `terminate()` doc all look good. Leaving M4 as-is is fine — browser parity, and observability can be a separate change.

One remaining gate is on my side, not this PR: I still need to run the full internal downstream JS suite before importing, because adding the `MessageChannel` global flips jasmine's `clearStack()` path (`setImmediate` → `MessageChannel` → `setTimeout`) for the internal suites and public CI doesn't cover those. That's an import concern, not a code concern.

clholgat:
Pulling in for more testing

clholgat (approved):

PR #134, opened 2026-08-02
Fix inspect command packaging

clholgat:
## Summary

- move the snapshot helper spec out of the runtime command source tree
- clean the CLI dist directory before compilation to prevent stale command files
- prevent published CLI builds from loading Jasmine specs as inspect commands

## Verification

- npm test: 91 specs, 0 failures
- npm run build
- verified dist/commands contains no spec or test JavaScript files
- ran the rebuilt CLI against a live Android app: inspect status connected and inspect devices found com.snap.valdi.ayab_valdi_app

### DIFF
--- a/npm_modules/cli/package.json
+++ b/npm_modules/cli/package.json
@@ -22,7 +22,7 @@
   ],
   "scripts": {
     "prepare": "npm run build",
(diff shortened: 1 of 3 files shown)


beaucollins (approved):
Accepting.

Starting internal acceptance process.

PR #142, opened 2026-08-08
fix android http cancelation

tx3stn:
## Description

Android `DefaultHTTPRequestManager` diverges from the iOS/macOS defaults in two ways:

1. **Cancellation does nothing.** `HTTPRequestTask.cancel()` nulls the completion but never touches the connection, so the transfer runs to completion and the result is thrown away. 
iOS actually cancels the `NSURLSessionTask`, so it works there.

2. **Requests are serial.** The executor is `ThreadPoolExecutor(0, 1, …)`. iOS uses `NSURLSession.sharedSession`, which is concurrent by default.

So a cancelled request neither stops nor frees the only worker, and everything else queues behind it until it finishes on its own.

### Changes

- `cancel()` disconnects the connection (a request cancelled while queued never opens a connection).
- Pool sized to 4, matching `NSURLSession.HTTPMaximumConnectionsPerHost` (so Android behaviour is consistent with iOS)
- Added unit tests for these and evrything with `bazel test //valdi:test_java --test_output=errors` is passing.

## Type of Change
- [x] Performance optimization
- [ ] Test improvement
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
+++ b/valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
@@ -1,17 +1,53 @@
 package com.snap.valdi.network
 
 import android.content.Context
-import com.snap.valdi.utils.ExecutorsUtil
+import java.io.IOException
 import java.io.InputStream
 import java.net.HttpURLConnection
 import java.net.URL
-import java.util.concurrent.Executors
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
 import com.snapchat.client.valdi.*
 import com.snapchat.client.valdi_core.*
 
 class DefaultHTTPRequestManager(context: Context): HTTPRequestManager() {
 
     private class RequestTask(val url: URL, val method: String, val body: ByteArray?, val headers: Map<String, String>, completion: HTTPRequestManagerCompletion): HTTPRequestTask(completion), Runnable {
+
+        private var connection: HttpURLConnection? = null
+        private var cancelled = false
+
+        override fun cancel() {
+            super.cancel()
+
+            val connectionToClose = synchronized(this) {
+                cancelled = true
+                connection
+            }
+
+            // Closing from another thread is what makes the worker's blocked read throw, which is
+            // how the thread gets reclaimed. Do it outside the lock so a slow close cannot stall
+            // the task binding its connection.
+            connectionToClose?.disconnect()
+        }
+
+        private fun isCancelled(): Boolean = synchronized(this) { cancelled }
+
+        /**
+         * Hands the connection to [cancel] so it can be torn down. Returns false when the request
+         * was already cancelled, in which case the caller must not go on to perform it.
+         */
+        private fun bindConnection(urlConnection: HttpURLConnection): Boolean = synchronized(this) {
+            if (cancelled) {
+                false
+            } else {
+                connection = urlConnection
+                true
+            }
+        }
 
         private fun doPerformRequestWithURLConnection(urlConnection: HttpURLConnection): HTTPResponse {
             urlConnection.instanceFollowRedirects = true
@@ -29,6 +65,13 @@ class DefaultHTTPRequestManager(context: Context): HTTPRequestManager() {
                     urlConnection.doOutput = true
                     urlConnection.outputStream.write(body)
                     urlConnection.outputStream.close()
+                }
+
+                // Reading responseCode is what opens the socket, and disconnect() cannot tear down
+                // a connection that has not connected yet. Checking here keeps a cancel that
+                // arrived during setup from starting a request nobody is waiting for.
+                if (isCancelled()) {
+                    throw IOException("Request was cancelled")
                 }
 
                 val responseCode = urlConnection.responseCode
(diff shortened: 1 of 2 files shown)


tx3stn:
failing job appears to be a timeout, not related to my changes, but I don't have permissions to re-run.

valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
```diff
@@ -1,17 +1,60 @@
...
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
 import com.snapchat.client.valdi.*
 import com.snapchat.client.valdi_core.*
 
-class DefaultHTTPRequestManager(context: Context): HTTPRequestManager() {
+class DefaultHTTPRequestManager(
+    context: Context,
+    private val openConnection: (URL) -> URLConnection = { it.openConnection() },
```
afan-sc: It looks like adding this defaulted Kotlin parameter removes the existing `DefaultHTTPRequestManager(Context)` constructor from the published JVM ABI.

I double-checked the compiled before/after classes with javap. The PR artifact only has (Context, Function1) and the synthetic default-argument constructor. a reflection test for (Context) fails with `NoSuchMethodException`. The in-repo Kotlin caller recompiles successfully, but external Java callers and binaries linked against the existing constructor can break.

Would it make sense to preserve the public Context constructor and keep the injectable connection factory behind an internal test seam?

valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
```diff
@@ -93,22 +156,35 @@ class DefaultHTTPRequestManager(context: Context): HTTPRequestManager() {
...
+                return RequestTask(url, method, body, headers, openConnection, completion)
             }
         }
     }
 
-    private var executors = ExecutorsUtil.newSingleThreadCachedExecutor { r ->
+    private val threadCount = AtomicInteger(0)
+
+    private val executors: ExecutorService = ThreadPoolExecutor(
+        MAX_CONCURRENT_REQUESTS,
```
afan-sc: Seems like this limits Android to four requests globally, whereas `URLSession`’s connection limit is per host.

I double-checked with four stalled requests to `127.0.0.1` followed by an immediately responding request using the distinct hostname `localhost`. The latter remained queued because all four executor workers were occupied. This means one slow host can still block unrelated hosts, unlike the `HTTPMaximumConnectionsPerHost` behavior referenced in the PR description.

Is the global cap intentional? If so, could we clarify the parity claim? Otherwise, would it make sense to preserve capacity across hosts?

valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
tx3stn: good catch! gated behind `VisibleForTesting`.
ran `javap -p -cp bazel-bin/valdi/valdi_java_kt-kt.jar com.snap.valdi.network.DefaultHTTPRequestManager | grep -i "DefaultHTTPRequestManager(\|DefaultConstructorMarker"`

and get the following:
```
  public com.snap.valdi.network.DefaultHTTPRequestManager(android.content.Context, kotlin.jvm.functions.Function1<? super java.net.URL, ? extends java.net.URLConnection>);
  public com.snap.valdi.network.DefaultHTTPRequestManager(android.content.Context);
```

valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
tx3stn: yeah that's my mis-read, sorry.

the connections per host difference means this is an android improvement, but not ios parity.
I looked into a few options to get something closer to ios logic. Looks like the best way if having them more closely aligned would be to use https://github.com/lysine-dev/okhttp as the default (but that's a discussion way beyond the scope of this fix - _side note, I did check to see if I could set okhttp as the default in my app without any Valdi changes required via `ValdiRuntimeManager.addRequestManager`, but ran into a bug there - I'll raise a fix for that separately_).

So, I've made a small update with tests to support the same maximum requests per hosts behaviour, with tests to the cross paltform behaviour aligns better.

valdi/src/java/com/snap/valdi/network/DefaultHTTPRequestManager.kt
```diff
@@ -93,22 +160,42 @@ class DefaultHTTPRequestManager(context: Context): HTTPRequestManager() {
...
 
-    private var executors = ExecutorsUtil.newSingleThreadCachedExecutor { r ->
-        Thread(r).apply {
-            name = "Valdi Network Thread"
-            priority = Thread.NORM_PRIORITY
+    private val threadCount = AtomicInteger(0)
+
+    // One pool per host, matching NSURLSession.httpMaximumConnectionsPerHost, so a stalled host
+    // cannot starve requests to other hosts.
+    private val executors = ConcurrentHashMap<String, ExecutorService>()
```
afan-sc: The per-host pools address the cross-host starvation concern, thanks for adding the test! One non-blocking question, should we separately track bounding the number of host pools or evicting them when idle, since the executor map currently retains an entry for every hostname?

tx3stn: Pools time out idle threads on a 60s keep-alive. I've added a small addition so now each request drops any pool left threadless, so the map holds hosts active in the last 60s rather than every hostname seen, which should mitigate most of that.

Bounding the total possible number does seem like a sensible thing to do, happy to look into that as it's own separate PR to stop this one getting too big.

PR #144, opened 2026-08-12
fix: Validate project name as a Bazel module name during bootstrap

ecoreng:
## Problem

`valdi bootstrap` accepts a project name like `MyProject`, then fails at the very end, after the project files are already written:

```
Error in module: invalid module name 'MyProject': valid names must
1) only contain lowercase letters (a-z), digits (0-9), dots (.), hyphens (-),
and underscores (_); 2) begin with a lowercase letter; ...
```

The name becomes `module(name = ...)` in `MODULE.bazel`, which Bazel requires to be lowercase. Nothing checks that until Bazel runs. Re-running bootstrap then fails with `Detected existing project files`, so you have to clean up by hand first (or the cleanup version)

## Fix

Check the name against Bazel's rules at the prompt, before any files are created:

```
❌ Project name "MyProject" is not a valid Bazel module name. It must only contain
lowercase letters (a-z), digits (0-9), dots (.), hyphens (-) and underscores (_),
begin with a lowercase letter and end with a lowercase letter or digit.
Did you mean "myproject"?
```

The name is not corrected automatically — directory and target names stay case-sensitive, so that stays the user's choice.

Note: `123project` is now rejected instead of being turned into `_123project`.

### DIFF
--- a/npm_modules/cli/src/utils/stringUtils.ts
+++ b/npm_modules/cli/src/utils/stringUtils.ts
@@ -80,7 +80,7 @@ export interface ValidateProjectNameOptions {
    * `valdi bootstrap`). Module and target names created by `valdi new_module`
    * are case-sensitive directory/BUILD target names and do not need this.
    */
-  bazelModule?: boolean;
+  requireBazelModuleName?: boolean;
 }
 
 /**
@@ -104,7 +104,7 @@ export function validateProjectName(name: string, options: ValidateProjectNameOp
 
   // When the name is used as the Bazel module name in MODULE.bazel, reject anything
   // Bazel would refuse before any files are written.
-  if (options.bazelModule && !isValidBazelModuleName(sanitized)) {
+  if (options.requireBazelModuleName && !isValidBazelModuleName(sanitized)) {
     const suggestion = sanitized.toLowerCase().replace(/^[^a-z]+/, '').replace(/[^\da-z]+$/, '');
     return (
       `Project name "${name}" is not a valid Bazel module name. ` +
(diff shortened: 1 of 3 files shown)


ecoreng:
If we really consider this an error it should fail quickly like here.

If we want to "sanitize" it we can but then we're assuming it's ok to change iMac to imac and it may not be brand friendly or whatever. Regardless there is a `// Warn if the name was significantly changed during sanitization` and casing change may also be this.

npm_modules/cli/src/utils/stringUtils.ts
```diff
@@ -79,7 +91,19 @@ export function validateProjectName(name: string): string | null {
   if (RESERVED_PROJECT_NAMES.has(sanitized.toLowerCase())) {
     return `Project name "${name}" (sanitized to "${sanitized}") is a reserved word and cannot be used. Please choose a different name.`;
   }
-  
+
+  // The name is used as the Bazel module name in MODULE.bazel, so reject anything
+  // Bazel would refuse before any files are written.
+  if (!isValidBazelModuleName(sanitized)) {
```
afan-sc: It looks like this Bazel-specific validation is being added to the shared validateProjectName(), which is also used by valdi new_module in newModule.ts:163

I double-checked the existing smoke-test name testNewModule: the new regex rejects it, so the command throws before creating the module. CamelCase is valid for these directory and BUILD target names because they do not become the root module(name = ...).

Would it make sense to apply isValidBazelModuleName() only in the bootstrap path, or split project and Valdi-module validation?

PR #146, opened 2026-08-13
test(companion): pin @ExportModule first-root-node detection

beaucollins:
## What

Adds a regression test for `dumpRootNodes` (companion `AST.ts`) covering how a file is detected as an `@ExportModule` module.

## Why

`dumpRootNodes` decides `shouldDumpAllExportedSymbols` by inspecting **only `rootNodes[0]`** (`hasExportModuleAnnotation(rootNodes[0])`). So the same `@ExportModule` annotation produces different dumped output depending on where in the file it appears — a whole-file scan would treat the file as an export module regardless of position. This test pins that first-node-only behavior so any change to it is deliberate.

## Test

`compiler/companion/src/AST.spec.ts` — one case, two placements:
- `@ExportModule` leading the **first** root node → the plain exported interface is dumped.
- `@ExportModule` leading a **later** root node → the plain exported interface is dropped.

Verified locally against the companion jest suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/compiler/companion/src/AST.spec.ts
+++ b/compiler/companion/src/AST.spec.ts
@@ -75,6 +75,55 @@ describe('AST', () => {
     expect(firstMemberOfComponent.leadingComments?.text).toContain('@Action');
   });
 
+  // `dumpRootNodes` decides whether a file is an @ExportModule module by inspecting only
+  // rootNodes[0] (the first top-level node). So the same @ExportModule annotation changes the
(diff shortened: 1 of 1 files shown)


beaucollins:
Closing — the @ExportModule first-node detection concern is being dismissed, so this behavior-pinning test isn't needed.

PR #148, opened 2026-08-18
Valdi Web sync

scorsin-oai:
## Description
This change syncs up most of the Valdi web implementation in the OpenAI fork back to upstream.

- Adds Valdi Web as a complete supported platform, including DOM rendering, layout, styling, text and inputs, images, animations, navigation, gestures, custom views, workers, persistence, and browser lifecycle handling.
- Introduces valdi install web, giving developers a first-class command to build and launch a Valdi application in the browser. Web also participates in the existing hot-reload workflow.
- Adds Bazel support for Web applications and reusable npm packages, including web_deps, native-module registration, worker entrypoints, configurable asset inlining, tree shaking, and package metadata.
- Extends the compiler for Web module generation and loading, lazy module resolution, native-module shims, path collapsing, inline localization, worker registries, and reliable hot-reload source mapping.
- Aligns Web typography with host applications by inheriting font settings and smoothing and mapping Valdi system font weights appropriately.
- Adds the cross-platform integration-test module: 137 renderer cases, deterministic snapshot capture, node-output comparison, native image diffing, and an interactive HTML report for comparing Web and native renderers.
Includes a framework-level SSR prototype with a server DOM, HTML/CSS serialization, routing, streaming responses, and tests.

## Type of Change
- [ ] Test improvement
- [ ] Other (please describe)

## Testing

## Checklist


clholgat:
Pulled `08a94ae` into the public [AYAB Valdi update](https://github.com/clholgat/ayab-valdi/pull/3) and tested it against another downstream Web consumer.

AYAB builds, all 11 Bazel test targets and 17 browser E2E workflows pass, and simulation works. The second consumer builds and its canonical Valdi/package verification passes with compatibility patches. These are the issues I hit:

### 1. External paths escape the collapsed output tree

With Valdi consumed through Bzlmod, `_dest_native()` can receive paths like:

```text
../valdi~/src/valdi_modules/src/valdi/coreutils/src/RuntimeBase.d.ts
```

The fallback returns that unchanged, so the collapse action writes outside its declared tree artifact.

[Patch used in AYAB](https://github.com/clholgat/ayab-valdi/blob/273aece42e541c36501b056716f410780f865909/src/patches/valdi/0002-Normalize-external-web-native-paths.patch)

### 2. Current Widgets still uses `setColorPalette`

`InitSemanticColors.ts` calls `runtime.setColorPalette()`, which this branch removes in favor of named palettes.

[Widgets compatibility patch](https://github.com/clholgat/ayab-valdi/blob/273aece42e541c36501b056716f410780f865909/src/patches/valdi_widgets/0002-Use-named-color-palette-runtime-API.patch)

The matching Widgets change needs to land with this, or Valdi needs a temporary compatibility wrapper.

### 3. PersistentStore changes format without migrating existing data

The previous Web implementation stored:

```text
valdi.PersistentStore.<store-name>
```

The new implementation stores individual entries under:

```text
valdi.persistence.v1:<scope>:<store-name>:<entry-key>
```

I couldn't find a fallback reader or migration. Existing preferences and recoverable app state disappear after upgrading.

### 4. Native-module shim generation and ownership are fragile

I expected `@ExportModule` plus the Web implementation in `web_deps` to be sufficient. Instead:

- Without `web_register_native_module_id_overrides`, normal module imports remain unresolved.
- Overrides declared by dependencies aren't reliably propagated into downstream collapsed packages.
- Declaring an override on an owning module can make the concrete output and generated shim target the same URL, causing a duplicate `.finalFile` error.

[Temporary AYAB override](https://github.com/clholgat/ayab-valdi/blob/273aece42e541c36501b056716f410780f865909/src/modules/process_image/BUILD.bazel)

This needs one ownership model, transitive aggregation, and automatic inference from the exported declaration and Web implementation rather than downstream registration metadata. The duplicate-output diagnostic should also identify the generated shim.

### 5. The browser-history navigation API has no equivalent replacement

`WebNavStack`, `RouteRegistry`, and the previous `WebNavigator` exposed URL construction, deep-link restoration, `popstate`, and back/forward navigation.

The new navigation implementation doesn't appear to expose equivalent browser-routing behavior. We currently have to restore the compatibility files downstream.

### 6. Generated `.bin` entries need undocumented Webpack configuration

`_module_entry_registry.js` emits static `require()` calls for `.bin` entries. A regular Webpack 5 consumer fails to parse them unless it adds:

```js
{
  test: /\.bin$/,
  type: 'asset/bytes',
}
```

Either the generated package needs to be bundler-neutral, or the required loader configuration needs to be shipped and documented.

### 7. Webpack reports a dynamic-require warning

Every build emits:

```text
Critical dependency: require function is used in a way in which dependencies
cannot be statically extracted
```

This appears to come from passing the raw `require` into dynamic component resolution. A generated static registry should avoid it.

### 8. Non-interactive layout overlays block pointer input

Catalog-style pages render normally, but cards underneath an empty absolute layout cannot be clicked. The new `LayoutElementClass` creates ordinary `<div>` elements without the previous `pointer-events: none` default, so an empty full-body layout wins `elementFromPoint()` and intercepts real pointer input.

The old contract was transparent layout/view nodes with interactive nodes opted back in by `onTap`, `touchEnabled`, or `hitTest`. Restoring that behavior fixes real mouse clicks through overlapping layouts. This needs a renderer test covering an empty overlay and an interactive child.

### 9. Pointer-transparent layouts also disable text fields and scroll views

After restoring transparent layout containers, the new `TextFieldElementClass` and `ScrollElementClass` remain `pointer-events: none`. Text fields cannot receive focus/input and wheel events cannot reach scroll views.

Hit-testing needs to be owned per element class: transparent layout overlays, pointer-active `onTap` views, editable text fields, and wheel/touch-scrollable scroll views should be tested together as one contract.

### 10. `onTouch` lost desktop mouse support and local coordinates

Widgets `Slider` uses `onTouch` as a cross-input drag primitive and calculates its value from element-local `event.x`. The previous renderer supplied a document-level mouse sequence and local `x`/`y`. This PR binds only browser touch events and reports viewport coordinates, so a real mouse drag leaves the slider unchanged.

The local fix restores desktop mouse handling, listener cleanup, and local coordinates while retaining `absoluteX`/`absoluteY`, but it has not been rerun yet. This should probably use Pointer Events plus pointer capture, with mouse, touch, drag-outside-bounds, coordinate, and cleanup tests.

clholgat:
Follow-up after another round of real-browser testing against `08a94ae`.

The corrected [input/scroll/mouse-event patch](https://github.com/clholgat/ayab-valdi/blob/5300af9e0c21eba9294dd2da7074b233b1303a74/src/patches/valdi/0010-Restore-web-input-scroll-and-mouse-touch-events.patch) now applies against the exact PR head. Keyboard and wheel tests pass, and a mouse drag changes the Slider zoom from 1.0x to 2.5x.

### 11. Multiline `TextView` is read-only by default

The rewritten `TextViewElementClass` emits a pointer-inactive `<div contentEditable="false">`. Multiline editors render but cannot receive focus or input.

[This patch](https://github.com/clholgat/ayab-valdi/blob/5300af9e0c21eba9294dd2da7074b233b1303a74/src/patches/valdi/0011-Restore-editable-Web-text-views.patch) restores editable behavior, including textbox accessibility attributes, while retaining `enabled={false}` as the explicit read-only state. A downstream browser workflow can enter multiline text and complete after the patch.

### 12. Static CommonJS imports bypass the selected Web implementation

This is a follow-up to item 4. Some compiled consumers contain imports such as:

```js
require("../../../raster_image/src/RasterImage.js")
```

Webpack resolves that import directly to the native implementation. The Web implementation registered by generated `RegisterNativeModules` code is only selected when the runtime module loader is involved. The resulting chart layout is present, but no image or canvas is rendered.

A downstream `NormalModuleReplacementPlugin` mapping proves that selecting the generated Web implementation fixes the workflow, but this should not require downstream registration or bundler replacement. The collapsed Web package/generated imports should resolve the Web implementation directly.

An end-to-end fixture where the native and Web implementations produce observably different output would cover this path.

### 13. Labels inherit `flex-shrink: 0`

`createBaseLayoutItemElement()` sets `flexShrink: 0`, and rewritten labels inherit it. Long status text therefore expands its row and causes horizontal page overflow at narrow viewport widths.

[This patch](https://github.com/clholgat/ayab-valdi/blob/5300af9e0c21eba9294dd2da7074b233b1303a74/src/patches/valdi/0013-Restore-shrinkable-web-labels.patch) restores the previous shrinkable label default and adds a renderer regression test. The downstream narrow-viewport test then has no surrounding page overflow, while its intentional chart scroller still works.

The current downstream patch set passes the upstream Web renderer test, 11 project tests, and 17 browser workflows. [Full notes and repro details are here.](https://github.com/clholgat/ayab-valdi/blob/5300af9e0c21eba9294dd2da7074b233b1303a74/docs/valdi-web-sync-pr-148-notes.md)

scorsin-oai:
@clholgat Thanks for the detailed downstream testing and repros. I followed up on both rounds of feedback:

1. Fixed external/Bzlmod path normalization in `4684e767`.
2. Restored `runtime.setColorPalette()` as a deprecated compatibility wrapper in `69c89f0d`, so the current public Widgets release remains buildable.
3. Restored the existing public Web persistence implementation and `valdi.PersistentStore.<name>` storage format in `745975b7`. Since the format no longer changes, existing browser data does not need a migration.
4. The explicit Bazel mapping remains intentional. `web_deps` contributes build inputs but does not define logical native-module IDs. `web_register_native_module_id_overrides` is owned by the module providing the Web implementation and is collected transitively by the collapse rule. It is meant to bind declaration-only modules to Web implementations, not to replace a concrete JavaScript implementation at the same module ID.
5. Restored `WebNavStack`, `RouteRegistry`, browser history, deep-link restoration, and `popstate` navigation in `1702686f`.
6. Kept direct `.bin` imports and documented the required Webpack 5 `asset/bytes` rule in `937e2f8a`.
7. Removed the dynamic-require warning in `89038632`: component resolution now goes through the runtime's generated component registry instead of passing raw `require`. The integration Webpack build no longer emits the critical-dependency warning.
8–11. Fixed pointer-transparent layout overlays, restored pointer activity for controls and scroll views, moved `onTouch` to pointer-based document-level dragging with element-local coordinates, and made multiline TextViews editable unless `enabled={false}` in `03d884cd`. The focused renderer coverage exercises overlays, interactive children, inputs, scrolling, pointer dragging, coordinates, cleanup, and TextView editability.
12. No replacement behavior was added here because Web overrides are specifically for a declaration-only `.d.ts` module backed by a Web implementation. They are not intended to replace an existing compiled JavaScript file at the same module ID. A static import of a concrete JavaScript implementation therefore continues to resolve that implementation directly.
13. Fixed in `745e8f38`. Labels now use the browser's shrinkable flex-item behavior, while preserving native Valdi semantics: an unset `numberOfLines` behaves as `1`, and only `numberOfLines={0}` enables unlimited wrapping. The Web renderer suite passes, and a 138-case before/after snapshot run verifies that a default label stays on one constrained line while the multiline label wraps inside the same row. A repeat capture produced an identical targeted snapshot.

PR #154, opened 2026-08-25
Add browser-based Valdi debugger

bjdodson-openai:
## Description

Adds a packaged, browser-based Valdi debugger available through `valdi debugger`.

The debugger provides:

- Live daemon target discovery and attachment
- Native view hierarchy and inert HTML preview
- Component ViewModel/state inspection
- Root and element snapshots
- Heap report export and runtime log streaming
- Hermes CPU profiling
- External action/state synchronization for automation

Component debug data is explicitly opt-in, bounded, and excluded from existing `valdi inspect tree` requests by default. The local server is restricted to loopback and validates Host, Origin, request size, action names, ports, and projected media sources.

This also adds CLI packaging, documentation, CI path coverage, and regression tests.

## Type of Change

- [x] Documentation improvement
- [x] Test improvement
- [x] Other (new debugger feature)

## Testing

### Testing Details

- `npm test`: 136 specs passed
- CLI build and package dry run passed
- `bazel query //...` passed
- The focused Valdi core target builds successfully, and all 17 new debugger serialization tests pass. The aggregate target remains red in the unchanged `ModuleLoader.preloadBatch` suite.
- Manually started the built debugger server and verified:
  - JSON startup output and graceful shutdown
  - Packaged UI and security headers
  - Loopback/Host/Origin protections
  - Invalid action and port rejection
  - No local filesystem paths exposed through debugger status or log metadata

## Checklist

## Related Issues

N/A

## Additional Context

This PR intentionally establishes the native debugger foundation only.

Deferred follow-ups:

- Web inspection should land together with its first-party web bridge.
- Renderer tracing requires the larger runtime/native instrumentation port first.
- Input forwarding and data/network provider tooling should land with their corresponding runtime contracts and end-to-end tests.

### DIFF
--- a/npm_modules/cli/debugger/index.html
+++ b/npm_modules/cli/debugger/index.html
@@ -1,0 +1,219 @@
+<!doctype html>
+<html lang="en" data-theme="sunset">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Valdi Debugger</title>
+    <link rel="stylesheet" href="./debugger.css" />
+  </head>
+  <body>
+    <div class="app">
+      <header class="topbar">
+        <div class="title-row">
+          <div class="brand-mark">V</div>
+          <div class="title-copy">
+            <h1>Valdi Debugger</h1>
+            <div class="subtitle" id="sessionSubtitle">No session selected</div>
+          </div>
+        </div>
+        <div class="toolbar">
+          <span class="status-pill"
+            ><span class="dot good" id="reloaderDot"></span><span id="reloaderState">Hot reloader online</span></span
+          >
+          <span class="status-pill"
+            ><span class="dot warn" id="proxyDot"></span><span id="proxyState">Proxy 9010 idle</span></span
+          >
+          <details class="connection-drawer">
+            <summary class="connection-summary">
+              <span class="connection-summary-text">Targets / Daemon</span>
+              <span class="connection-summary-hint">Details</span>
+            </summary>
+            <div class="connection-content">
+              <section class="connection-section">
+                <div class="connection-section-header">
+                  <h2 class="panel-title">Targets</h2>
+                  <button class="icon" id="refreshTargets" title="Refresh targets">R</button>
+                </div>
+                <div class="target-list" id="targetList"></div>
+              </section>
+              <section class="connection-section">
+                <h2 class="panel-title">Daemon Status</h2>
+                <div class="daemon-status" id="daemonStatus"></div>
+              </section>
+            </div>
+          </details>
+          <select id="portSelect" title="Preview renderer or Valdi daemon target">
+            <option value="13591">SnapDrawing / standalone :13591</option>
+            <option value="13592">mobile :13592</option>
+          </select>
+          <button id="attachButton" class="primary">Attach</button>
+          <button id="reloadButton" title="Fetch the latest Valdi tree and preview snapshot">Refresh</button>
+          <label title="Track live host tree, scroll offsets, and preview snapshot"
+            ><input id="autoRefreshToggle" type="checkbox" />Auto</label
+          >
+        </div>
+      </header>
+
+      <main class="debugger-shell">
+        <nav class="section-rail" aria-label="Debugger sections">
+          <button class="section-button active" data-section-button="ui">
+            <strong>UI</strong>
+            <span>Preview, hierarchy, inspector</span>
+          </button>
+          <button class="section-button" data-section-button="performance">
+            <strong>Performance</strong>
+            <span>Hermes CPU profiles</span>
+          </button>
+          <button class="section-button" data-section-button="logs">
+            <strong>Logs</strong>
+            <span>Runtime logs, commands</span>
+          </button>
+        </nav>
+
+        <section class="section-panel active" data-section-panel="ui">
+          <div class="workspace">
+            <section class="main">
+              <div class="export-panel" id="exportPanel">
+                <div class="export-header">
+                  <div class="export-title" id="exportTitle">Export</div>
+                  <div class="export-actions">
+                    <a id="exportOpenLink" target="_blank" rel="noopener">Open JSON</a>
+                    <button id="copyExportButton">Copy JSON</button>
+                    <button id="closeExportButton">Close</button>
+                  </div>
+                </div>
+                <textarea id="exportText" spellcheck="false" readonly></textarea>
+              </div>
+
+              <div class="main-toolbar">
+                <div class="main-title">
+                  <h2 class="panel-title">Live Preview / View Hierarchy</h2>
+                  <span id="hierarchySource" class="source-pill">Empty</span>
+                </div>
+                <div class="main-toolbar-actions">
+                  <div class="segmented" aria-label="Preview mode">
+                    <button id="modeLive" class="active">Live</button>
+                    <button id="modeIssues">Issues</button>
+                  </div>
+                  <button
+                    id="copyPreviewButton"
+                    title="Copy the live preview image, or snapshot JSON if no image is available"
+                  >
+                    Copy
+                  </button>
+                  <div class="status-pill">
+                    <span class="dot good"></span><span id="selectedSummary">No node selected</span>
+                  </div>
+                </div>
+              </div>
+
+              <div class="stage" id="previewStage">
+                <div class="device" id="deviceFrame">
+                  <div class="screen" id="screen">
+                    <div class="app-preview" id="appPreview">
+                      <div id="htmlPreviewRoot" class="html-preview-root" aria-label="Live Valdi HTML preview"></div>
+                      <img id="rootSnapshotImage" class="root-snapshot" alt="" hidden />
+                    </div>
+                  </div>
+                </div>
+              </div>
+
+              <div class="hierarchy-dock" aria-label="View hierarchy">
+                <div class="tree-controls">
+                  <input id="treeSearch" type="search" placeholder="Filter tags, ids, attributes" />
+                </div>
+                <div class="tree" id="tree"></div>
+              </div>
+            </section>
+
+            <aside class="panel inspector">
+              <div class="panel-header">
+                <h2 class="panel-title">Inspector</h2>
+                <div>
+                  <button class="icon" id="snapshotButton" title="Capture selected element snapshot">S</button>
+                  <button class="icon" id="copyPathButton" title="Copy selected path">C</button>
+                </div>
+              </div>
+              <div class="tabs">
+                <button data-tab="overview" class="active">Overview</button>
+                <button data-tab="props">Props</button>
+                <button data-tab="state">State</button>
+                <button data-tab="issues">Issues</button>
+                <button data-tab="raw">Raw</button>
+              </div>
+              <div class="inspector-content" id="inspector"></div>
+            </aside>
+          </div>
+        </section>
+
+        <section
+          class="section-panel performance-panel"
+          data-section-panel="performance"
+          aria-label="Performance tools"
+        >
+          <div class="section-heading">
+            <h2>Performance</h2>
+            <p>Capture Hermes CPU profiles for a connected Valdi JavaScript context.</p>
+          </div>
+          <div class="performance-grid">
+            <details class="performance-card" open>
+              <summary class="performance-card-header">
+                <div class="performance-title">CPU</div>
+                <span class="source-pill" id="profileStatusPill">Idle</span>
+              </summary>
+              <div class="performance-controls">
+                <select id="profileContextSelect" title="Hermes profile context">
+                  <option value="">Auto context</option>
+                </select>
+                <button id="profileRefreshButton" title="Refresh Hermes contexts">Contexts</button>
+                <label
+                  ><input id="profileDurationInput" type="number" min="0.1" max="60" step="0.5" value="5" />s</label
+                >
+                <button id="profileStartButton">Start</button>
+                <button id="profileStopButton">Stop</button>
+                <button id="profileCaptureButton" class="primary">Capture</button>
+                <button id="profileExportButton">Export</button>
+              </div>
+              <div class="performance-summary" id="profileSummary">No CPU profile captured.</div>
+            </details>
+          </div>
+        </section>
+
+        <section class="section-panel" data-section-panel="logs">
+          <div class="console">
+            <div class="console-top">
+              <div class="console-title">Runtime Logs</div>
+              <div class="log-filter">
+                <input id="logSearch" type="search" placeholder="Filter logs" />
+              </div>
+              <button id="clearLogsButton" class="console-action" title="Clear logs shown in this debugger session">
+                Clear logs
+              </button>
+            </div>
+            <div class="log-list" id="logs"></div>
+            <form class="console-prompt" id="commandForm">
+              <span class="prompt-glyph">&gt;</span>
+              <input
+                id="commandInput"
+                autocomplete="off"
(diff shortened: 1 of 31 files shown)


npm_modules/cli/src/debugger/server.ts
```diff
@@ -0,0 +1,1565 @@
...
+
+function portName(port: number): string {
+  if (port === STANDALONE_PORT) return 'standalone';
+  if (port === MOBILE_PORT) return 'mobile';
+  return 'custom';
+}
+
+function errorPayload(error: unknown): { error: string } {
+  return {
+    error: error instanceof Error ? error.message : String(error),
```
clholgat, line 318: `errorPayload` returns `error.message` verbatim, and the `handleApi` catch (line 1378) ships it to the client. Node `fs` errors embed absolute paths — e.g. `ENOENT: no such file or directory, scandir '/Users/<name>/.../logs'` — reachable via `/api/runtime-logs`, `/api/snapshot`, `/api/heap` when the daemon or logs dir is missing. That contradicts the PR's "No local filesystem paths exposed" testing note. Suggest returning a generic message for filesystem/`ENOENT`-class errors and logging the detail server-side only.

npm_modules/cli/debugger/index.html
```diff
@@ -0,0 +1,219 @@
+<!doctype html>
+<html lang="en" data-theme="sunset">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Valdi Debugger</title>
+    <link rel="stylesheet" href="./debugger.css" />
```
clholgat, line 7: No CSP is set on the served UI, and the (potentially untrusted) app view tree is projected into this top-level document rather than a sandboxed iframe. The client is well-escaped today, so this is defense-in-depth — but a strict CSP is nearly free here since every script is an external file:

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self'; object-src 'none'; base-uri 'none'; img-src 'self' data: blob:" />
```

Worth also considering a `sandbox`ed iframe for the live preview, since this UI drives a privileged local server (`/api/heap`, `/api/debugger/actions`).

src/valdi_modules/src/valdi/valdi_core/test/IRenderedVirtualNodeData.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+  it('does not expose component values through the compatibility serializer', () => {
+    const data = fromRenderedVirtualNode(
+      createComponentNode({ token: 'not-exported' }, { secret: 'not-exported' }),
+      true,
+    );
+
+    expect(data.component).toEqual({});
+  });
+
+  it('applies one component debug-data character budget across the tree', () => {
```
clholgat, line 243: Good coverage of the field cap, tree budget, and cycles. Two bounding limits the "bounded" claim rests on aren't exercised yet: the depth cap (`MAX_COMPONENT_DEBUG_DEPTH`, emitting `'...'` at max depth) and the per-collection 50-item cap (`MAX_COMPONENT_DEBUG_COLLECTION_ITEMS`, emitting the `more item(s)`/`more properties` marker for arrays/objects/maps/sets). Worth a test for each so those caps can't silently regress.

clholgat (commented):
Took a pass with a focus on the loopback server, the browser client, and the shared `valdi_core` serialization. Overall this is a solid, well-engineered PR — the security perimeter (loopback bind, Host allowlist, Origin + Sec-Fetch-Site CSRF, path-traversal containment, streaming body-size cap) is above average, the browser client is well-escaped with no live XSS, and the component-debug serialization is genuinely bounded and zero-cost on the non-debug path. A few small things worth fixing before merge, inline. No blockers.

npm_modules/cli/debugger/index.html
bjdodson-openai: Thanks for calling this out. The served UI is already covered by an HTTP response-header CSP in serveStatic: script-src self, connect-src self, frame-src none, object-src none, base-uri none, and frame-ancestors none. The server test also asserts the CSP plus X-Frame-Options DENY. A second meta policy would be redundant and, because policies intersect, the proposed version would drop the unsafe-inline style allowance required for the preview’s programmatic element styles. The preview is an inert projection rather than app HTML: it builds an allowlisted DOM with createElement, writes content through textContent/value, omits WebViews, and permits only data/blob media. Moving it into an iframe would require a cross-document selection/sizing channel and changing frame-src none, so I’m keeping the existing header CSP plus inert projection here. We can revisit an iframe boundary if the debugger ever renders app-authored HTML.

clholgat: You're right — I missed that `serveStatic` already sets this as a response header (`default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; …; frame-ancestors 'none'`) with `X-Frame-Options: DENY` and a test covering it. A `<meta>` policy would be redundant, and since policies intersect it could only tighten things in ways you don't want (e.g. dropping `style-src 'unsafe-inline'`). Agreed the inert createElement/textContent projection with data/blob-only media is a reasonable boundary, and deferring an iframe until app-authored HTML is rendered makes sense. Resolving — thanks for the detail.

npm_modules/cli/src/debugger/server.ts
clholgat: Confirmed — the `isFileSystemError` guard + `clientErrorPayload` (generic client message, real error to `console.warn`) is exactly right, and good call applying it to `streamRuntimeLogs`, `collectClientContexts`, and `inspectPort` too, not just the `handleApi` catch. The test asserting the sanitized response *and* the preserved server-side log is a nice touch. Resolving.

src/valdi_modules/src/valdi/valdi_core/test/IRenderedVirtualNodeData.spec.ts
clholgat: These are exactly the two I was after — the depth cap (`level4: ...`) and the 50-item array/object caps with their omission markers, covering both collection shapes. Thanks for adding them. Resolving.

clholgat:
Thanks @bjdodson-openai — all review feedback is addressed and the fixes look good. I'm pulling this into our internal monorepo now for further internal testing. (Heads up: the remaining red C++ Tests check is a pre-existing flaky death test — `AsyncStrictModeSyncCallAssertsOnMainThread` hanging on `fork()` in a threaded context and timing out — unrelated to these changes.) I'll follow up here with anything internal testing surfaces.

npm_modules/cli/src/debugger/server.ts
bjdodson-openai: Addressed in 1145e94e. `clientErrorPayload` now keeps filesystem details in the server-side `console.warn` while returning a generic client message, and it is used by the API catch, runtime-log stream, context collection, and port inspection paths. `server.spec.ts` verifies the response omits the absolute logs directory while the warning retains it.

src/valdi_modules/src/valdi/valdi_core/test/IRenderedVirtualNodeData.spec.ts
bjdodson-openai: Addressed in 1145e94e. `IRenderedVirtualNodeData.spec.ts` now verifies the depth cap emits `level4: ...` without the nested value, and the 50-item cap emits omission markers for both arrays and objects while excluding item/property 50.

npm_modules/cli/src/debugger/server.ts
```diff
@@ -0,0 +1,1581 @@
...
+  }
+}
+
+function createDebuggerHttpServer(host: string, port: number): Server {
+  return http.createServer(async (request, response) => {
+    const url = new URL(request.url || '/', `http://${hostForUrl(host)}:${port}`);
+    const apiFetchSiteAllowed =
+      !url.pathname.startsWith('/api/') || isAllowedApiFetchSite(request.headers['sec-fetch-site']);
+    if (
+      !isAllowedRequestHost(request.headers.host, port) ||
```
clholgat, line 1554: 🔴 High: The `/api/*` surface has no authentication — loopback binding plus these Host/Origin/Sec-Fetch-Site checks are the only gate. A request with no `Origin` and no `Sec-Fetch-Site` header passes all three: `isAllowedRequestOrigin` returns true when Origin is undefined, and `isAllowedApiFetchSite` returns true when the header is undefined. Failure scenario: any other local user or process on the host runs `curl http://127.0.0.1:8765/api/snapshot` and gets live component viewModel/state, or POSTs `/api/heap` for a full JS heap and reads `/api/runtime-logs` — data that can contain tokens/PII. Loopback is not a per-user trust boundary. Suggest binding a random per-session token (in the served URL path or a required header) and rejecting `/api/*` requests that lack it, similar to the Chrome DevTools ws GUID.

npm_modules/cli/src/debugger/server.ts
```diff
@@ -0,0 +1,1581 @@
...
+  MOBILE_PORT,
+  type RemoteContext,
+  STANDALONE_PORT,
+  connectToDaemon,
+} from '../utils/daemonClient';
+import { getUserConfig, resolveFilePath } from '../utils/fileUtils';
+import { type CpuProfile, HERMES_PORT, HermesConnection, listHermesDevices } from '../utils/hermesClient';
+
+const DEFAULT_HOST = process.env['VALDI_DEBUGGER_HOST'] || '127.0.0.1';
+const DEFAULT_PORT = Number.parseInt(process.env['VALDI_DEBUGGER_PORT'] || '8765', 10);
```
clholgat: 🔴 High (cross-PR): `VALDI_DEBUGGER_PORT` is overloaded. Here it sets the browser UI HTTP port (default 8765); in the runtime (PR #155, `resolveDebuggerPortWithDiagnostics`) the same variable sets the on-device debugger TCP service port (default 13591/13592). On top of that, this server's discovery hardcodes `STANDALONE_PORT`/`MOBILE_PORT` (13591/13592) in `inspectStatus`/`inspectSnapshot` and never reads the env var. Failure scenario: `export VALDI_DEBUGGER_PORT=14000; <launch app>; valdi debugger` → the UI mis-binds to 14000 instead of 8765 and still probes 13591/13592, so it never finds the app now on 14000. Suggest splitting the names (e.g. `VALDI_DEBUGGER_UI_PORT` vs `VALDI_DEBUGGER_SERVICE_PORT`) and letting discovery honor the configured service port.

clholgat (commented):
Focused review of the debugger foundation (`server.ts`). The security mechanics are largely solid: loopback-only bind (rejects 0.0.0.0/non-loopback, handles IPv6 ::1), DNS-rebind-resistant Host/Origin validation, path-traversal-safe static serving, `data:`/`blob:`-only preview media sources, and consistent `escapeHtml` on all client render paths (no XSS found). Two items stand out for the trust model and the port contract, plus a couple of correctness notes; see inline comments.

Note: since this 24-PR stack will be squashed into ~3 PRs, please carry this feedback into the squashed PRs.

Low-severity:
- 🟢 Low — `summarizeCpuProfile` echoes `node.callFrame.url` verbatim in `topFunctions`. Hermes frame URLs can be local bundle paths (`file://…`), which would contradict the "no filesystem paths exposed" goal for the profiling path. Worth confirming against the bundle URL form and stripping/relativizing if so.

npm_modules/cli/src/debugger/server.ts
```diff
@@ -0,0 +1,1581 @@
...
+  server: Server;
+  close: () => Promise<void>;
+  host: string;
+  port: number;
+  url: string;
+  requestedPort: number;
+  portWasAutoSelected: boolean;
+}
+
+const devEventClients = new Set<ServerResponse>();
```
clholgat, line 130: 🟠 Med: Request/session state lives in module-level globals (`activeHost`, `assetRoot`, `activeLogsDirectory`, `debuggerUiState`, `devEventClients`/`debuggerEventClients`, `activeProfileSession`). A second `startDebuggerServer()` in the same process overwrites this shared state, and all connected clients mutate one shared `debuggerUiState` with no isolation. Since this is the base of the stack, suggest encapsulating per-server state in the object returned by `startDebuggerServer` before more features build on these globals.

npm_modules/cli/src/debugger/server.ts
```diff
@@ -0,0 +1,1581 @@
...
+): Promise<Record<string, unknown>> {
+  if (request.method !== 'POST') {
+    throw new Error('CPU profile capture requires POST.');
+  }
+  return await runProfileTransition(async () => {
+    assertNoActiveProfile();
+    const body = await readJsonBody(request);
+    const durationMs = clampNumber(readBodyNumber(body, 'durationMs', 5000), 100, 60_000);
+    activeProfileSession = await createProfileSession(searchParams, body);
+    await delay(durationMs);
```
clholgat: 🟠 Med: `captureCpuProfile` holds the `profileTransitionInProgress` lock across this `await delay(durationMs)` (clamped up to 60s), so `stopCpuProfile`/`startCpuProfile`/`captureCpuProfile` all throw "Another CPU profile transition is already in progress" for the entire capture window. Failure scenario: user starts a 60s capture, clicks Stop → 400, and cannot abort until it finishes; it also delays graceful shutdown. Suggest holding the lock only around session setup/teardown (not the sampling delay) and letting stop interrupt an active capture.

npm_modules/cli/src/debugger/server.ts
bjdodson-openai: Addressed in 5169cef. Each debugger server now generates a fresh 32-byte cryptographically random token. Every /api route requires it in the X-Valdi-Debugger-Token header; only the two EventSource routes accept the same token as a query parameter because the browser EventSource API cannot set headers. The served no-store HTML receives the token bootstrap and sends Referrer-Policy: no-referrer. Coverage verifies that missing and incorrect header/query tokens receive 401, a query token cannot authenticate an ordinary API route, and authenticated POST and SSE requests succeed.

npm_modules/cli/src/debugger/server.ts
bjdodson-openai: Addressed in 5169cef. The browser UI now reads VALDI_DEBUGGER_UI_PORT, while target discovery reads VALDI_DEBUGGER_SERVICE_PORT. The configured service port becomes the discovery default instead of probing only the hard-coded standalone/mobile pair. Tests set the two variables to distinct values, verify strict validation, and assert that status discovery probes and reports the configured service port. The downstream native port slice is also being replaced by the consolidated debugger commit using the service-port name.

npm_modules/cli/src/debugger/server.ts
bjdodson-openai: Addressed in 5169cef by making the CLI contract explicit: one debugger server may run per process. A concurrent second start now fails deterministically instead of overwriting shared state; close is idempotent and resets the lifecycle guard, streams, watchers, profile state, target settings, and UI state so a later server starts cleanly. A regression test covers rejection of the second server and successful restart after close. The shared debuggerUiState is intentional collaboration state for clients of that single local server; independent multi-server isolation is not a supported CLI mode.

npm_modules/cli/src/debugger/server.ts
bjdodson-openai: Addressed in 5169cef. Timed capture now holds the transition lock only for profile setup and teardown, with the sampling interval represented by an interruptible operation. Stop interrupts an active timed capture and shares the same single completion; shutdown also stops an outstanding capture before closing. The regression test starts a 60-second operation, stops it immediately, and verifies both callers receive the same result with exactly one completion.

PR #161, opened 2026-08-26
Add bounded debugger performance tracing

bjdodson-openai:
## Description

Adds lifecycle-safe, bounded performance tracing across the CLI, renderer, runtime, and native tracer layers.

- Preserves the existing native tracer ABI while adding bounded debugger results.
- Handles uncertain Start/Stop outcomes, reload recovery, target disappearance, and result expiry.
- Caps full response envelopes, trace events, dropped-event counts, and retained results.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (new debugger capability)

## Testing
- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [ ] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details
- Incremental branch: Restacked CLI 261/261 and production build passed; focused server/daemon coverage and independent native ABI review passed.
- Assembled debugger stack: `npm test` passed 436/436; the CLI production build passed.
- Focused `//src/valdi_modules/src/valdi/web_renderer:test` passed.
- `bazel query //...` passed.
- The broad Valdi suite reproduced the established 12 unrelated failures; all new debugger specs passed.

## Checklist
- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues
Relates to #154

## Additional Context
Stack 8/22. Stacked on #160 (`bjd/debugger-settings`). Review this PR as the single incremental commit `06be7132` against that base; do not merge it before its parent.

PR #162, opened 2026-08-26
Add generic debugger data providers

bjdodson-openai:
## Description

Adds one bounded provider/settings protocol with strict target identity and truthful unavailable states.

- Uses serialized-document contracts instead of exposing arbitrary provider objects.
- Generation-isolates discovery and actions across target replacement and module reload.
- Bounds request and response envelopes, descriptor access, settings identity, and concurrency.

## Type of Change
- [x] Test improvement

## Testing

### Testing Details
- Incremental branch: CLI 281/281 and production build passed; focused provider/settings, adversarial bounds, reload, native unload, and browser syntax checks passed.
- `bazel query //...` passed.

## Checklist

## Related Issues
Relates to #154

## Additional Context
Stack 9/22. Stacked on #161 (`bjd/debugger-tracing`). Review this PR as the single incremental commit `0cc8ec48` against that base; do not merge it before its parent.

### DIFF
--- a/src/valdi_modules/src/valdi/valdi_core/src/debugging/DebuggerProvider.ts
+++ b/src/valdi_modules/src/valdi/valdi_core/src/debugging/DebuggerProvider.ts
@@ -1,0 +1,918 @@
+import { jsx } from '../JSXBootstrap';
+import { getModuleLoader } from '../ModuleLoaderGlobal';
+import type { ValdiRuntime } from '../ValdiRuntime';
+import type { CustomMessageHandler } from './CustomMessageHandler';
+
+declare const module: { readonly path?: string };
+declare const runtime: ValdiRuntime;
+
+const DEBUGGER_PROVIDERS_IDENTIFIER = 'ValdiDebuggerProviders';
+const DEBUGGER_PROVIDERS_CONTRACT_VERSION = 1;
+const MAX_DEBUGGER_CUSTOM_RESPONSE_BYTES = 128 * 1024;
+const MAX_DEBUGGER_ACTION_JSON_BYTES = 48 * 1024;
+const MAX_DEBUGGER_JSON_DEPTH = 8;
+const MAX_DEBUGGER_JSON_VALUES = 10_000;
+const MAX_DEBUGGER_JSON_COLLECTION_ITEMS = 100;
+const MAX_DEBUGGER_JSON_PROPERTY_NAME_CHARACTERS = 1024;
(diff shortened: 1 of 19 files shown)


src/valdi_modules/src/valdi/valdi_core/src/debugging/DebuggerProvider.ts
```diff
@@ -0,0 +1,918 @@
...
+  const capturedGeneration = bridge.generation;
+  const capturedRevision = bridge.revision;
+  const capturedToken = provider.token;
+  provider.inFlight++;
+  try {
+    const result = await provider.handleRequest(request);
+    if (
+      provider.disposed ||
+      bridge.generation !== capturedGeneration ||
+      bridge.revision !== capturedRevision ||
```
clholgat, line 740: 🟠 Med — staleness guard rejects valid results on *unrelated* registry changes. `capturedRevision` is snapshotted before `await provider.handleRequest(request)`, and any later bump to `bridge.revision` fails this check and returns `{stale:true}`. But `revision` is incremented by *any* provider's register/dispose/`notifyChange()`, not just this one — and this provider's own identity is already covered by the `provider.disposed` / `capturedGeneration` / `capturedToken` checks on the surrounding lines.

Failure scenario: a slow Storage `snapshot` is in flight when a second adapter registers (or an adapter calls the documented `notifyChange()` on a data change) → the completed, valid snapshot is discarded, the client maps `stale` → `null`, and the Data panel renders empty instead of the data.

This is the shared provider contract the ClientSQL fork (#176/#177) builds on, so a fork that registers lazily or fires `notifyChange()` during live use will hit it.

Suggested fix: drop `bridge.revision !== capturedRevision` from the condition (rely on disposed/generation/token), or recompute `base` from the current registry rather than discarding the data.

clholgat (commented):
Please carry this feedback into the squashed PR(s).

Also (🟢 Low): `debugger-providers.js` `debuggerProviderForKind()` collapses providers to one-per-kind via `.find(p => p.kind === kind)`, while the core contract allows many providers keyed by `id` (and `list` returns all, sorted by kind/label). If two providers ever share a kind (e.g. the ClientSQL fork's `sql` provider alongside another `sql`-kind provider), only the first by sort order is selectable and the other is silently unreachable in the UI. Worth a per-provider selection surface, or documenting kind as effectively unique in the client.

bjdodson-openai:

PR #167, opened 2026-08-26
Add bounded web performance capture API

bjdodson-openai:
## Description

Adds a dormant Chromium recording controller and bounded loopback API for web performance capture.

- Enforces one exact global CDP tracing owner and fail-closed uncertain termination.
- Normalizes trace output under response, event, string, and lifecycle bounds.
- Separates the capture API from the later presentation layer.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (new debugger capability)

## Testing
- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [ ] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details
- Incremental branch: Recorder/server 89/89, full CLI 331/331, production build, and repository query passed; the isolated exact-next-port rerun and subsequent suites passed.
- Assembled debugger stack: `npm test` passed 436/436; the CLI production build passed.
- Focused `//src/valdi_modules/src/valdi/web_renderer:test` passed.
- `bazel query //...` passed.
- The broad Valdi suite reproduced the established 12 unrelated failures; all new debugger specs passed.

## Checklist
- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues
Relates to #154

## Additional Context
Stack 14/22. Stacked on #166 (`bjd/debugger-web-tracing-compat`). Review this PR as the single incremental commit `f7ad9a8a` against that base; do not merge it before its parent.

PR #168, opened 2026-08-26
Add the web preview performance panel

bjdodson-openai:
## Description

Adds a generation-safe Performance panel over the preceding bounded capture API.

- Keeps Start, Stop, polling, and completed results bound to one exact target generation.
- Preserves focus and scroll while rejecting stale or ambiguous capture state.
- Cleans up ownership on target changes, aborts, and page lifecycle events.

## Type of Change
- [x] Test improvement

## Testing

### Testing Details
- Incremental branch: Panel 25/25, combined focused 114/114, full CLI 347/347, production build, TypeScript, lint, format, syntax, and query passed.
- `bazel query //...` passed.

## Checklist

## Related Issues
Relates to #154

## Additional Context
Stack 15/22. Stacked on #167 (`bjd/debugger-web-performance-api`). Review this PR as the single incremental commit `c1d9db42` against that base; do not merge it before its parent.

### DIFF
--- a/npm_modules/cli/debugger/devtools-panel.js
+++ b/npm_modules/cli/debugger/devtools-panel.js
@@ -4,6 +4,9 @@ const inspectedTargetNonce = query.get('targetNonce');
 const MAX_CONSOLE_ENTRIES = 500;
 const MAX_CONSOLE_ENTRY_CHARACTERS = 50_000;
 const MAX_CONSOLE_HISTORY_ENTRIES = 100;
+const MAX_PERFORMANCE_SAMPLES = 120;
+const MAX_PERFORMANCE_TIMELINE_ROWS = 120;
+const MAX_PERFORMANCE_SUMMARY_ROWS = 12;
 
 const state = {
   target: null,
@@ -31,6 +34,23 @@ const state = {
   consoleHistoryIndex: 0,
   consoleStream: null,
   consoleStreamTargetKey: null,
+  performance: {
+    data: null,
+    durationSeconds: 3,
+    error: null,
+    lastTrace: null,
+    navigationExpanded: false,
+    operationGeneration: 0,
+    ownerIdentity: null,
+    pending: false,
+    rendererTracingEnabled: false,
+    requestGeneration: 0,
+    samples: [],
+    snapshotPending: false,
+    traceActive: false,
+    traceScope: 'valdi',
+    traceSearch: '',
+  },
   error: null,
 };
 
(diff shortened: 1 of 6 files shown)


npm_modules/cli/debugger/devtools-panel.js
```diff
@@ -105,6 +139,79 @@ function formatNumber(value) {
...
+}
+
+function formatUptime(value) {
+  const numeric = Number(value);
+  if (!Number.isFinite(numeric) || numeric < 0) return '—';
+  if (numeric < 60_000) return `${formatNumber(numeric / 1000)} s`;
+  return `${formatNumber(numeric / 60_000)} min`;
+}
+
+function performanceIdentity(target = state.target) {
```
clholgat, line 165: 🟠 Med — the frontend "exact identity" guarantee is effectively sessionId-only. `inspectedUrl` and `inspectedTargetNonce` are module-level constants read once from the panel URL, so `performanceIdentity()` rebuilds them identically on every call and `performanceIdentityIsCurrent()` can never observe an `inspectedUrl`/`targetNonce` mismatch — only `sessionId` can ever differ.

Failure scenario: the panel advertises that Start/Stop/polling/results are bound to the exact `inspectedUrl` and per-tab `targetNonce` and "fail closed", but that enforcement lives entirely server-side. If a stale/rebound tab shares a `sessionId`, the client-side guard would not catch it; safety currently rests solely on #167's server-side nonce validation. Either derive the identity from the actual live target (so the fields can genuinely differ) or drop/soften the client-side guarantee in the docs so it isn't mistaken for real cross-check.

clholgat (commented):
Review of the web-preview Performance panel. Rendering is safe: every dynamic value (trace names, metric values, labels, search box, duration, paint names, counts) is routed through `escapeHtml` before `innerHTML`, and `preparePerformanceForTargetChange` nulls `data`/`lastTrace`/`samples` on every target change, so no XSS and no cross-generation data bleed. Polling is guarded by `pending`/`snapshotPending`, skips when `document.hidden`, and stops on pagehide. One identity concern inline, plus two low notes below.

Note: please carry this feedback into the squashed PR when the stack is collapsed.

- 🟢 Low — one-shot Capture sets `traceActive`/`ownerIdentity` before issuing the request and does not reset them if the request throws. After a failed Capture the panel is wedged in "Result pending" / Stop-only until the next ~1200 ms poll recomputes owner state from `/status` — and only while the user stays on the Performance tab. Consider resetting owner state in a `catch` on the capture path.
- 🟢 Low — the `completionError` branch issues its trace/stop call against the *currently selected* identity, not necessarily `perf.ownerIdentity`. If the owner belongs to a previous target, the stop clears the wrong session's server state. Local owner state is guarded, but the network call is misdirected; route the stop to `perf.ownerIdentity` when present.

bjdodson-openai:

PR #172, opened 2026-08-26
Add bounded component-property inspection

bjdodson-openai:
## Description

Adds bounded read-only component-property inspection for exact web targets.

- Reads only own enumerable data descriptors and never invokes getters.
- Shares a 64 KiB emitted-value budget with bounded depth, keys, strings, and entries.
- Fails closed on hostile or revoked Proxies while preserving the hierarchy fallback.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (new debugger capability)

## Testing
- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [ ] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details
- Incremental branch: Focused registry/panel/server 149/149 and full CLI 398/398 passed; web-renderer Bazel and query passed, with the broad suite matching its exact baseline.
- Assembled debugger stack: `npm test` passed 436/436; the CLI production build passed.
- Focused `//src/valdi_modules/src/valdi/web_renderer:test` passed.
- `bazel query //...` passed.
- The broad Valdi suite reproduced the established 12 unrelated failures; all new debugger specs passed.

## Checklist
- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues
Relates to #154

## Additional Context
Stack 19/22. Stacked on #171 (`bjd/debugger-target-picker`). Review this PR as the single incremental commit `012775d6` against that base; do not merge it before its parent.

PR #173, opened 2026-08-26
Add the exact component-property edit protocol

bjdodson-openai:
## Description

Adds web-only, revision-bound scalar mutation authorization and a descriptor-stable renderer update path.

- Issues 128-bit, single-use tokens bound to exact target, component, property, type, and value revisions.
- Rejects stale identity, accessors, functions, inherited or forbidden properties, and non-finite values.
- Uses a shadow overlay that preserves source descriptors, prototypes, receiver semantics, and frozen or sealed invariants.

## Type of Change
- [x] Test improvement

## Testing

### Testing Details
- Incremental branch: Protocol boundary passed CLI 402/402 and production build; frozen-patch independent reviews, TypeScript, lint, format, syntax, and diff checks passed.
- `bazel query //...` passed.

## Checklist

## Related Issues
Relates to #154

## Additional Context
Stack 20/22. Stacked on #172 (`bjd/debugger-component-properties`). Review this PR as the single incremental commit `2dd01e3c` against that base; do not merge it before its parent.

### DIFF
--- a/npm_modules/cli/src/debugger/server.ts
+++ b/npm_modules/cli/src/debugger/server.ts
@@ -56,10 +56,19 @@ const PORT_SEARCH_LIMIT = 50;
 const MAX_RUNTIME_LOG_READ_BYTES = 1024 * 1024;
 const FATAL_JSON_UTF8_DECODER = new TextDecoder('utf8', { fatal: true });
 const WEB_PREVIEW_NONCE_PATTERN = /^[\w-]{16,128}$/;
+const COMPONENT_PROPERTY_TOKEN_PATTERN = /^[\da-f]{32}$/;
+const COMPONENT_PROPERTY_EDIT_ERROR = 'The component property edit is stale or invalid.';
+const MAX_COMPONENT_PROPERTY_NAME_CHARACTERS = 256;
+const MAX_COMPONENT_PROPERTY_STRING_BYTES = 65_536;
+const MAX_COMPONENT_ID_CHARACTERS = 4096;
+const FORBIDDEN_COMPONENT_PROPERTY_NAMES = new Set(['__proto__', 'children', 'constructor', 'prototype']);
 const MAX_DEBUGGER_TREE_NODES = 25_000;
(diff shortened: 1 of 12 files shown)


npm_modules/cli/src/debugger/server.ts
```diff
@@ -1542,18 +1551,130 @@ async function inspectWebPreviewSnapshot(searchParams: URLSearchParams): Promise
...
+    'componentToken',
+    'inspectedUrl',
+    'propertyName',
+    'sessionId',
+    'snapshotRevision',
+    'targetNonce',
+    'value',
+  ];
+  const keys = Object.keys(body).sort();
+  if (keys.length !== expectedKeys.length || keys.some((key, index) => key !== expectedKeys[index])) {
```
clholgat, line 1600: 🟠 Med — the edit wire contract is validated by exact key-set equality (and the bridge's `parseComponentPropertyEditRequest` does the same) with no protocol version field. But the bridge lives inside the inspected app (which can be any Valdi version), while the server/client are the independently-versioned CLI. Any future addition to the request tuple becomes a lock-step-only change with no negotiation: a newer CLI editing an app built with an older bridge (or vice-versa) will fail closed on the strict key-set check with no way to degrade gracefully. Failure scenario: a later release adds an optional field (e.g. an idempotency key) → every cross-version edit silently 400s/409s even though both sides support editing. Suggested fix: add an explicit `protocolVersion` field to the tuple now (accepted by both server and bridge), and validate a known-version key set per version rather than one frozen exact set, so the contract can evolve without a coordinated bridge+CLI upgrade.

clholgat (commented):
Review of the component-property edit protocol. The write path is well-bounded overall (single-use 128-bit tokens bound to target/component/property/type/revision, three-layer type+range validation, descriptor-stability double-capture, scalar-only, native targets structurally read-only). One design concern noted inline about the wire contract's forward/backward compatibility.

Note for the squash: please carry this feedback into the squashed PR that contains the edit protocol.

bjdodson-openai:

PR #180, opened 2026-08-28
Add cross-platform debugger DevTools capabilities

bjdodson-openai:
## Description

Collapses the reviewed post-foundation debugger stack into one dependency-complete commit atop current `main`.

This adds the cross-platform debugger capability set built across #155–#175: explicit ports and authenticated sessions, input control, the legacy web bridge, integrated DevTools, runtime settings, bounded storage and performance inspection, generic and persistent-storage providers, Chromium console streaming, web tracing and performance capture, console clearing, target discovery and selection, component hierarchy/property inspection and editing, and bounded runtime-state inspection.

The aggregate preserves the fixes and discussion outcomes from the original stack, including lifecycle and authentication repairs. Synthetic Slack-shaped redaction fixtures are assembled at runtime so no complete token signature is stored in the commit.

This PR supersedes #155, #156, #157, #158, #159, #160, #161, #162, #163, #164, #165, #166, #167, #168, #169, #170, #171, #172, #173, #174, and #175. Those PRs retain the incremental review history.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (feature: cross-platform debugger capabilities)

## Testing

- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [x] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details

- CLI Jasmine suite: 450/450 passing
- CLI production build: passing
- `bazel query //...`: 9,700 targets resolved
- Focused Bazel validation passed for runtime debugger, macOS Objective-C, and web renderer targets
- `git diff --check` and complete-token scans: clean
- Independent aggregate and post-redaction audits: clean

The full repository-wide `bazel test //...` matrix was not run locally; CI remains the final platform matrix.

## Checklist

- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues

Relates to #154.

Supersedes #155–#175 while preserving their review history.

## Additional Context

The foundation from #154 landed through Snap's internal import as `17754a1d`. This aggregate is one commit whose parent is the current upstream `main` commit and includes the reviewed post-foundation fixes omitted by that import.

PR #177, opened 2026-08-26
Add the portable ClientSQL runtime and debugger

bjdodson-openai:
## Description

Adds the complete portable ClientSQL runtime, native and web implementations, SQLite toolchain validation, generated integration, and debugger provider as one coherent native feature.

- Preserves bundle-scoped database identity, transaction ownership, hot-reload disposal, and bounded debugger payloads.
- Pins and validates the supported SQLite dialect and generated native integration.
- Keeps this optional side stack independent from the core debugger chain.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (new ClientSQL capability)

## Testing
- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [ ] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details
- Generator coverage passed 19/19.
- `//src/valdi_modules/src/valdi/client_sql:client_sql_native_tests` passed.
- `//valdi:test_client_sql_runtime_integration` passed.
- Focused SQLite validator, ClientSQL module, generated smoke bundle, and native-desktop smoke builds passed.
- Repository-wide `bazel query //...` completed successfully with 9,688 targets.

### Native API finalization

This PR introduces a new exported native API and intentionally carries six
`@Version(__PLACEHOLDER__)` annotations. During Snap's internal import/finalization
process, these declarations must be assigned one concrete runtime API version.

There are seven PR-added placeholder occurrences to finalize together: the six
annotations and the corresponding generator-test expectation. Existing
placeholder infrastructure and fixtures elsewhere in the repository must remain
unchanged.

The public Valdi repository does not currently expose a Native API Finalization
workflow, so this PR must not be merged with these declarations unresolved.
Please coordinate the version allocation with the internal importer/maintainer.

## Checklist
- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues
Relates to #154

## Additional Context
[...]

PR #176, opened 2026-08-26
Add reviewable ClientSQL generator sources

bjdodson-openai:
## Description

Adds reviewable and hermetic ClientSQL generator sources plus compiler integration, without exposing the native runtime.

- Uses one generated-symbol registry and comment-aware parameter lexing.
- Declares cross-package generator inputs so source and executed toolchains stay reproducible.
- Adds a real hermetic Bazel test target for generated output.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation improvement
- [ ] Performance optimization
- [x] Test improvement
- [x] Other (new ClientSQL capability)

## Testing
- [ ] Tests pass locally (`bazel test //...`)
- [x] Added/updated tests for changes (if applicable)
- [ ] Tested on multiple platforms (iOS/Android/Web/macOS as applicable)
- [ ] Manual testing performed (describe below)

### Testing Details
- Hermetic ClientSQL generator Bazel target passed 9/9.
- Generator syntax, formatting, diff, and generated-symbol checks passed.
- The completed optional side stack later passed all 19 generator tests, native tests, runtime integration, focused builds, and repository-wide Bazel analysis.

## Checklist
- [x] Code follows project style guidelines
- [x] Documentation updated (if needed)
- [x] No breaking changes (or documented in description)
- [x] Commit messages follow [conventional format](../CONTRIBUTING.md#commit-messages)
- [x] No secrets, API keys, or internal URLs included

## Related Issues
Relates to #154

## Additional Context
Optional side stack from generic providers #162 (`bjd/debugger-providers`). This ClientSQL stack is not required by the core debugger chain. Review this PR as the single incremental commit `2a6c0df1` against that base.

PR #181, opened 2026-08-28
Add portable ClientSQL runtime and debugger

bjdodson-openai:
## Description

Adds a dependency-complete portable ClientSQL implementation on top of the debugger capabilities landed from #180.

The single commit includes the hermetic ClientSQL generator, pinned SQLite toolchain, native and web runtimes, integration coverage, debugger provider, and a cross-platform Ledger SQL demo. The demo exercises reactive account and ledger queries, atomic transfers, reset behavior, and deterministic stress batches on macOS, iOS, and Android.

This revision also incorporates the review hardening requested on this PR: debug-database registration is inert unless runtime debugging is enabled; debugger query results are bounded; native transactions have timeout and exception-safe rollback paths; database teardown is safe from the coordinator queue; and the SQL generator now covers parameterized types, joins, nested constraints, `WITHOUT ROWID`, reserved identifiers, quoted identifiers and aliases, and scalar subqueries.

The SQLite source-archive provenance and global build-rule changes require the existing sensitive-file/import review.

This PR supersedes #176 and #177; those PRs retain the incremental review history.

## Type of Change

- [x] Documentation improvement
- [x] Test improvement
- [x] Other (feature: portable ClientSQL runtime, debugger provider, and Ledger SQL demo)

## Testing

### Testing Details

- `bazel test //compiler/clientsql:test_clientsql` — 24 tests passed
- `bazel test //src/valdi_modules/src/valdi/client_sql:client_sql_native_tests //valdi:test_client_sql_runtime_integration` — passed
- `bazel build //src/valdi_modules/src/valdi/client_sql:client_sql_web //apps/ledger_sql_demo:ledger_sql_demo_ios` — passed
- `bazel build //apps/ledger_sql_demo:ledger_sql_demo_android --snap_flavor=platform_development --copt=-DANDROID_WITH_JNI --repo_env=VALDI_PLATFORM_DEPENDENCIES=android --define=client_repo_arm64=true --android_platforms=@snap_platforms//os:android_arm64` — passed
- Ledger SQL demo macOS build and launch — passed; hot reload connected to the running app and delivered changes
- Repository-wide `bazel query //...` — passed
- `git diff --check` — clean

The full repository-wide `bazel test //...` matrix was not run locally; CI remains the final platform matrix.

## Checklist

## Related Issues

Builds on the capabilities landing from #180.

Supersedes #176 and #177 while preserving their review history.

## Additional Context

The single head commit is directly parented on the imported #180 capabilities landing at `b83dc83ee3d3587c3a94a3634fed560df3b0ccdd`. Review the ClientSQL-only range `b83dc83ee3d3587c3a94a3634fed560df3b0ccdd..f3ce8a09d972dc54aeca945cb35d325d0727dc36`.


src/valdi_modules/src/valdi/client_sql/src/ClientSQLDebug.ts
```diff
@@ -0,0 +1,928 @@
...
+
+function releaseDebugProviderIfUnused(): void {
+  if (Object.keys(databasesById).length > 0 || !debugProviderRegistration) {
+    return;
+  }
+  debugProviderRegistration.dispose();
+  debugProviderRegistration = undefined;
+}
+
+export function registerClientSQLDebugDatabase(database: ClientSQLDebugDatabase): void {
```
clholgat, line 889: **Trust boundary — please confirm this is gated.**

Every generated database class calls `registerClientSQLDebugDatabase(...)` on open, so all live ClientSQL databases auto-register with the debugger provider. The provider itself is nicely bounded (only `list`/`table`, read-only, identifiers quoted, values bound, payloads truncated), but it grants read access to arbitrary row/schema data for any open DB.

Can you confirm this registration and the underlying debugger transport (#180) are gated to debug builds and/or require authentication? A short comment here documenting that trust boundary would help.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1710 @@
...
+    if (!statement) {
+        return statement.moveError();
+    }
+
+    auto bindResult = bindParameters(database, statement.value().get(), request.parameters);
+    if (!bindResult) {
+        return bindResult.moveError();
+    }
+
+    std::vector<Value> rows;
```
clholgat, line 575: **Unbounded result set (High).** Rows accumulate into this `std::vector` with no cap or streaming, so a `query()`/`queryOnWriter()` over an un-`LIMIT`ed `SELECT` materializes the entire result set in native memory before returning to JS. SQL text length and bind sizes are capped elsewhere, but output volume isn't — consider a configurable row/byte ceiling or a streaming path.

clholgat (commented):
Really nice work — this is careful and thorough. Generated code is injection-safe throughout (values bound via `?`, identifiers escaped, literals `json.dumps`'d), RAII is correct in the native layer, the SQLite provenance is documented and hash-pinned, and the tests genuinely compile + execute generated output rather than snapshotting strings. Agreed it should stay in draft until the `@Version(__PLACEHOLDER__)` annotations are finalized.

I've left inline comments. Grouping and priority:

**Blocking / please confirm**
- Debugger auto-registration exposes read access to open DBs — is it gated to debug builds / authenticated? (comment on `ClientSQLDebug.ts`)
- Native: unbounded result set (High), and two transaction-liveness gaps (Med) that can wedge the writer or `std::terminate`.

**Generator — silently-wrong generated *types*, not crashes** (worth fixing since they hit common SQL): parameterized column types, `SELECT *` across a JOIN, `WITHOUT ROWID` tables, and the `await` reserved word.

**Two items I couldn't anchor cleanly inline:**
1. *Teardown/GC path (Med).* On the drop-without-explicit-close path, the last queued task dropping the final coordinator reference destroys the dispatch queue (and runs `sqlite3_close`) from within a task running on that same queue — a potential self-join deadlock/crash. The explicit-close path looks fine; this is the GC path, and it depends on the dispatch-queue destructor semantics, so worth a look.
2. *Non-transactional `query()` during an open write transaction* reads the pre-transaction WAL snapshot (dispatched to the reader pool, not deferred like `execute()`). Defensible reader-pool semantics, but an easy caller foot-gun — worth documenting.

**Suggested tests:** three fixtures — a parameterized column type, `SELECT *` across a JOIN, and a `WITHOUT ROWID` table — would catch the top three generator findings directly.

Nothing here is a redesign; the structure is solid. Happy to dig deeper on any of these.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1710 @@
...
+        }
+
+        Error sqlError;
+        if (!execRawSQL(database.value(), "BEGIN TRANSACTION", &sqlError)) {
+            notifyClientSQLCallbackError(callback, sqlError);
+            return;
+        }
+
+        const auto transactionId = self->_nextTransactionId++;
+        self->_activeTransactionId = transactionId;
```
clholgat, line 1245: **Transaction liveness (Med).** If the JS transaction body returns without ever invoking its callback and without throwing (e.g. a dropped/never-settled promise), `_activeTransactionId` stays non-zero indefinitely — the `BEGIN` holds the write lock and all subsequent writer work is deferred forever. `busy_timeout` only covers lock contention, not this. Consider a native transaction watchdog/timeout.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1710 @@
...
+                self->finishTransaction(
+                    transactionId,
+                    callback,
+                    std::move(value),
+                    std::move(error));
+            };
+
+        try {
+            body(std::move(transaction), std::move(doneCallback));
+        } catch (const Exception& error) {
```
clholgat, line 1280: **Catch is too narrow (Med).** This only catches `const Exception&`. A transaction body that throws anything else escapes the writer-thread dispatch lambda uncaught (likely `std::terminate`) and leaves the `BEGIN` un-rolled-back — the same wedged state as the never-settled case. Suggest catching all throws here and rolling back on any of them.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1710 @@
...
+
+Result<Void> checkSQLiteStatus(sqlite3* database, int status, std::string_view operation) {
+    if (status == SQLITE_OK || status == SQLITE_DONE || status == SQLITE_ROW) {
+        return Void();
+    }
+
+    return Error(STRING_FORMAT("ClientSQLNative {} failed: {}", operation, clientSQLSQLiteError(database)));
+}
+
+bool execRawSQL(sqlite3* database, std::string_view sql, Error* outError) {
```
clholgat, line 82: **Nit (Low).** `sqlite3_exec` with `std::string(sql).c_str()` truncates at the first embedded NUL, so a schema/migration statement containing `\0` would be silently cut short. Only affects developer-authored schema SQL, so low risk, but worth guarding.

compiler/clientsql/src/clientsql/sql.py
```diff
@@ -0,0 +1,760 @@
...
+
+def infer_expression_type(expression: str) -> str:
+    if re.search(r"\bCOUNT\s*\(", expression, re.IGNORECASE):
+        return "number"
+    if re.search(r"\b(?:SUM|AVG|MIN|MAX)\s*\(", expression, re.IGNORECASE):
+        return "number | null"
+    return "any"
+
+
+def default_ts_type(sql_type: str) -> str:
```
clholgat, line 1000: **Parameterized column types collapse to `any` (Med-High).** `default_ts_type` only matches bare type names, and `clean_identifier` doesn't strip parens, so `sql_type` for `VARCHAR(255)` / `DECIMAL(10,2)` / `NUMERIC(18,4)` is the literal string and falls through to `any`. Compounding it, `parse_columns` (around line 217) tokenizes with `.split()`, so `price NUMERIC(10, 2) NOT NULL` (space after the comma) splits the type across tokens and yields `sql_type="NUMERIC(10,"`. These are very common declarations and each one silently loses type safety. No test currently covers a parameterized type.

compiler/clientsql/src/clientsql/sql.py
```diff
@@ -0,0 +1,760 @@
...
+    return fields
+
+
+def select_star_table(sql: str, tables: Dict[str, Table]) -> Optional[Table]:
+    if not re.search(r"\bSELECT\s+\*\s+FROM\b", sql, re.IGNORECASE | re.DOTALL):
+        return None
+    return table_for_query(sql, tables)
+
+
+def table_for_query(sql: str, tables: Dict[str, Table]) -> Optional[Table]:
```
clholgat, line 939: **`SELECT *` across a JOIN produces an incomplete Row type (Med).** `table_for_query` returns only the first `FROM` table, so `SELECT * FROM post JOIN author ...` generates a Row type with only `post` columns while the runtime rows contain the author columns too. The explicit-column JOIN path is also lossy (columns resolve against the first table only), but at least doesn't misname the type.

compiler/clientsql/src/clientsql/sql.py
```diff
@@ -0,0 +1,760 @@
...
+            continue
+        if current_key is None:
+            continue
+        type_match = re.search(r"\b(?:typescript|ts|type)\s*:\s*['\"]?([^'\"]+)['\"]?", line, re.IGNORECASE)
+        if type_match:
+            result[current_key] = type_match.group(1).strip()
+    return result
+
+
+def collect_tables(sql_texts: Iterable[str], custom_types: Dict[str, str]) -> Dict[str, Table]:
```
clholgat, line 70: **`CREATE TABLE ... WITHOUT ROWID;` not captured (Med).** The body-capture requires the closing `)` to be immediately followed by `;`, so a trailing table option (`) WITHOUT ROWID;`, valid on the 3.16 floor) means no match and the table is absent from type inference — its Row/params degrade to empty/`any` even though the DDL is still emitted and validated. Paren-bearing table-level constraints (`CHECK(...)`, composite keys) can truncate the captured column body similarly.

compiler/clientsql/src/clientsql/sql.py
```diff
@@ -0,0 +1,760 @@
...
+import sys
+from pathlib import Path
+from typing import Dict, Iterable, List, Optional, Sequence, Tuple
+
+from .model import Column, ParamOccurrence, Parameter, Query, SqlFile, Table, ClientSqlError
+
+
+SQL_IDENTIFIER_PATTERN = r"[`\"\[]?[A-Za-z_][A-Za-z0-9_]*[`\"\]]?"
+MAX_INT32 = 2_147_483_647
+TYPESCRIPT_RESERVED_IDENTIFIERS = {
```
clholgat, line 15: **Reserved-word set omits `await` (Low-Med).** `yield`/`let`/`static` are here, but not `await` (also `async`, `as`, `of`). A param named `:await` would emit `selectX(await: number)`, a SyntaxError in module/async scope. Query/param names come from the developer, so unusual but real.

compiler/clientsql/src/clientsql/sql.py
```diff
@@ -0,0 +1,760 @@
...
+    params = infer_params(sql, param_order, tables)
+    read_tables = tables_read_by_query(sql, tables)
+    changed_tables = tables_changed_by_query(sql, tables)
+    # Result-shape inference is deliberately conservative. SQLite validates more
+    # statement forms below, but generated row types are only promised for the
+    # SELECT grammar that infer_result_fields() understands.
+    returns_rows = bool(re.match(r"^\s*SELECT\b", sql, re.IGNORECASE))
+    result_fields = infer_result_fields(sql, tables) if returns_rows else []
+    result_type = f"{sanitize_type_name(name)}Row" if returns_rows else "void"
+    query_table = table_for_query(sql, tables)
```
clholgat: Nit: `query_table` is assigned here and never used — dead assignment.

MODULE.bazel
```diff
@@ -183,6 +183,22 @@ http_archive(
     url = "https://github.com/fmtlib/fmt/releases/download/7.1.3/fmt-7.1.3.zip",
 )
 
+http_archive(
+    name = "sqlite",
```
clholgat: Nit: this `http_archive` block is duplicated in `bzl/dependencies.bzl` for the WORKSPACE path. The consistency test helps catch drift, but a single shared definition would remove the hazard entirely.

src/valdi_modules/src/valdi/client_sql/src/ClientSQLDebug.ts
bjdodson-openai: Confirmed. The shared `registerDebuggerProvider()` path returns an inert registration whenever `runtime.isDebugEnabled` is false, so ClientSQL's automatic registration exposes no provider in non-debug runtimes. ClientSQL intentionally adds no second authentication layer; access remains governed by #180's host/build-gated debugger transport. I added a comment here documenting that boundary and the embedder requirement.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Fixed. The common native materializer now enforces both a 10,000-row ceiling and a conservative 8 MiB allocation estimate, including row/cell overhead, column labels, and scalar/text/blob payloads. This covers `query()`, `queryOnWriter()`, and `transaction.query()` because all three use `queryStatement()`. Exceeding either ceiling returns an error with no partial result. Covered by `boundsQueryResultRowMaterialization`, `boundsQueryOnWriterResultByteMaterialization`, and `countsNullCellAndColumnOverheadTowardQueryLimit`.

bjdodson-openai:
Addressed both unanchored findings as well:

1. On the drop-without-close path, `ClientSQLConnection` now abandons its handle explicitly. When the final coordinator reference is released from one of its owned queues, the coordinator transfers its queue-owning writer/reader references to the fallback release queue before the current task unwinds, avoiding self-destruction/self-join. `releasesFinalDroppedHandleWithoutClose` verifies that the coordinator returns to the baseline live count.
2. The reader-pool behavior is now documented under “Query snapshot semantics.” A non-transactional `query()` may observe the committed WAL snapshot available when its reader starts; it does not provide read-your-writes or writer callback ordering. The README directs callers to `transaction.query()` for transaction-local reads and `queryOnWriter()` for serialization behind writer work. `allowsSeparateHandleSnapshotReadersDuringActiveNativeTransaction` verifies the before/after-commit behavior.

I also added the suggested parameterized-type, joined-`SELECT *`, and `WITHOUT ROWID` generator fixtures.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Fixed with a native watchdog. Transactions default to a 30-second deadline; if the completion callback never arrives, the watchdog atomically claims completion, rolls back on the writer queue, clears the active transaction, and drains deferred writer work. Late or duplicate callbacks are ignored. The timeout is injectable for tests; `timesOutAndRollsBackTransactionThatNeverCompletes` verifies rollback, writer recovery, and single completion.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Fixed. The transaction boundary now handles Valdi `Exception`, `std::exception`, and unknown native throws. Every path atomically claims completion, cancels the watchdog, rolls back, and releases deferred writer work. `rollsBackWhenTransactionBodyThrowsStandardException` verifies the standard-exception path and subsequent database usability.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Fixed at the native API boundary. Every create statement and migration statement is checked for an embedded NUL before opening or mutating the database, so `sqlite3_exec` can no longer silently see a truncated statement. `rejectsEmbeddedNullInSchemaAndMigrationSQL` covers both paths.

compiler/clientsql/src/clientsql/sql.py
bjdodson-openai: Fixed. Column parsing now finds the balanced declaration body, separates constraints at top level, normalizes multi-token types, and removes type parameters only for affinity mapping. `VARCHAR(255)`, `NUMERIC(10, 2)`, and `DECIMAL(18, 4)` therefore retain their SQL declarations while generating `string`/`number` types. `test_preserves_schema_and_types_for_parameterized_joined_tables` covers these forms and compiles the generated TypeScript.

compiler/clientsql/src/clientsql/sql.py
bjdodson-openai: Fixed. Query inference now models every top-level `FROM`/`JOIN` source and alias, expands `*` in source order, resolves qualified explicit columns against the correct source, and applies nullable types to outer-joined sources. Tests cover `SELECT *`, explicit aliases, `LEFT JOIN`, conflicting joined-column parameter types, and generated watcher dependencies.

compiler/clientsql/src/clientsql/sql.py
bjdodson-openai: Fixed. `collect_tables()` now locates the opening parenthesis and scans to its balanced closing parenthesis, so nested `CHECK(...)` and composite constraints cannot truncate the body and trailing table options no longer prevent collection. The generator fixture includes two `WITHOUT ROWID` tables plus a nested `CHECK`, and verifies that the emitted DDL remains intact.

compiler/clientsql/src/clientsql/sql.py
bjdodson-openai: Fixed. Added `await`, `async`, `as`, and `of` to the TypeScript reserved set; generated parameter names are prefixed with `_`. The fixture verifies the generated signature and runs TypeScript compilation over the output.

bjdodson-openai: Fixed. Added `await`, `async`, `as`, and `of` to the TypeScript reserved set; generated parameter names are prefixed with `_`. The fixture verifies the generated signature and runs TypeScript compilation over the output.

compiler/clientsql/src/clientsql/sql.py
bjdodson-openai: Removed — thanks.

MODULE.bazel
bjdodson-openai: I looked at consolidating this. `MODULE.bazel` and the legacy WORKSPACE entry point cannot directly reuse the same repository-setup macro; moving the Bzlmod side behind a module extension would broaden this change considerably. I kept the declarations in their required entry points, added comments documenting why, and retained `test_native_sqlite_dependency_shape_is_explicit_for_apple_and_default` to pin both URLs and hashes and catch drift. Happy to pursue a module-extension cleanup separately if preferred.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1919 @@
...
+    _readerConnections.reserve(kClientSQLReaderConnectionCount);
+    for (size_t index = 0; index < kClientSQLReaderConnectionCount; ++index) {
+        auto readerQueue = createClientSQLQueue(STRING_LITERAL("Valdi ClientSQL Reader"), true, fallbackQueue);
+        if (readerQueue != nullptr) {
+            _readerConnections.emplace_back(makeShared<ClientSQLQueueConnection>(std::move(readerQueue)));
+        }
+    }
+}
+
+ClientSQLDatabaseCoordinator::~ClientSQLDatabaseCoordinator() {
```
clholgat, line 984: **Teardown/GC path (Med) — I don't think this case is covered yet.** The explicit `close()` path looks fine. But on the drop-without-explicit-close path, when the last queued writer/reader task drops the final coordinator reference, this destructor (and the `sqlite3_close` in the connection handle deleter) can run from *within a task executing on the coordinator's own dispatch queue* — destroying a threaded queue from its own thread is a classic self-join hazard. Worth verifying against the dispatch-queue destructor semantics, or steering the GC path through the writer-thread close.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
```diff
@@ -0,0 +1,1919 @@
...
+
+Result<Void> checkSQLiteStatus(sqlite3* database, int status, std::string_view operation) {
+    if (status == SQLITE_OK || status == SQLITE_DONE || status == SQLITE_ROW) {
+        return Void();
+    }
+
+    return Error(STRING_FORMAT("ClientSQLNative {} failed: {}", operation, clientSQLSQLiteError(database)));
+}
+
+bool execRawSQL(sqlite3* database, std::string_view sql, Error* outError) {
```
clholgat, line 82: Nit: `std::string(sql).c_str()` truncates at the first embedded NUL, so a schema/migration statement containing `\0` would be silently cut short. Developer-authored SQL only, so low risk.

src/valdi_modules/src/valdi/client_sql/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+        "//conditions:default": ["VALDI_CLIENTSQL_USE_BUNDLED_SQLITE"],
+    }),
+    visibility = ["//visibility:public"],
+    deps = [
+        ":client_sql_cpp",
+        "//valdi:valdi_runtime",
+    ] + select({
+        "//bzl/conditions:ios": [],
+        "//bzl/conditions:macos": [],
+        "//conditions:default": ["@sqlite//:sqlite"],
```
clholgat: Portability request: consider exposing the SQLite dependency as an overridable label (e.g. a `label_flag` defaulting to `@sqlite//:sqlite`) rather than hardcoding it here. That lets an embedder that already vendors SQLite point this at their copy and avoid linking a second amalgamation with duplicate `sqlite3_*` symbols — without patching this BUILD file. Same idea would help for the validator's `@sqlite_316`.

src/valdi_modules/src/valdi/client_sql/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+    }),
+)
+
+valdi_test(
+    name = "client_sql_native_tests",
```
clholgat, line 24: Gating: `client_sql_native_tests` (and `//valdi:test_client_sql_runtime_integration`) are standalone targets. Consider folding them into the `//valdi:test` aggregate so they run wherever `//valdi:test` does, instead of each CI config having to enumerate them separately.

valdi/BUILD.bazel
```diff
@@ -946,6 +947,30 @@ valdi_test(
     ],
 )
 
+# Runs the generated TestDb TypeScript through a real Valdi JS runtime while
+# linking ClientSQL's self-registering native module factory. This intentionally
+# remains a distinct target so the optional SQLite dependency does not expand
+# every Valdi integration test binary.
+valdi_test(
+    name = "test_client_sql_runtime_integration",
```
clholgat, line 952: (Companion to the gating note on `client_sql_native_tests`.) This runtime integration target sits outside the `//valdi:test` aggregate, so it won't run unless a CI list names it explicitly — folding it into `//valdi:test` would gate it everywhere.

clholgat (commented):
Thanks for the fast turnaround — this addresses the review well. Confirmed on `8e1b957`: result-set caps (row + byte, with a clear "add a LIMIT" error), the transaction watchdog plus broadened `Exception&` / `std::exception&` / `...` catches, the `isDebugEnabled` trust-boundary comment, and on the generator the parameterized column types, `WITHOUT ROWID`, JOIN handling, and reserved-word set all look fixed. 👍

Left inline comments for the two small remaining items (teardown/GC path, NUL nit) and a couple of portability/gating requests (overridable SQLite labels, folding the new tests into `//valdi:test`).

**One design question** that informs how this slots into existing setups: is the emitted TypeScript intended to be output-compatible with SQLDelight v1, or is it a fresh contract? i.e. would an existing SQLDelight-v1 `.sq` consumer regenerate to equivalent bindings, or should they expect behavioral differences to revalidate? Not a blocker — just want to understand the migration story.

Nice work overall.

bjdodson-openai:
Answering the compatibility question: this generator is a fresh Valdi-specific SQLDelight-style contract. It is not currently demonstrated to be output-compatible with SQLDelight v1 or Snap’s internal ClientSQL, so existing consumers should not regenerate against it without revalidation.

The current `f3ce8a09` revision makes the public implementation coexistence-safe at the build boundary:

1. It preserves the established `@valdi_toolchain//:sqldelight_compiler` seam and the exact legacy `-p` / `-c` / `-m` invocation contract.
2. The portable generator is the public default behind that seam, while an internal embedder can retain its existing compiler target.
3. Runtime SQLite, validation SQLite, the whole native implementation, and implementation-specific aggregate native tests are independently overrideable.
4. The public SQLite repositories use collision-resistant `valdi_clientsql_sqlite` names.

Those changes prevent the open-source import from implicitly replacing Snap’s compiler or linking two native factories. They do not by themselves prove generated-binding, native-name, debugger-ID, database-path, or migration compatibility.

The remaining integration gate is:

1. Compare the internal and portable generated TypeScript/native metadata and behavior using the same sanitized SQL fixture.
2. Validate the merged internal graph: query, aggregate test link, Android/iOS exports, exactly one factory registration, and legacy database migration behavior.
3. Choose either atomic convergence on one canonical implementation while preserving internal contracts and storage behavior, or a fully namespaced `client_sql_v2` coexistence lane.

Generated artifacts are sufficient; internal source is not needed. The useful inputs are the current ClientSQL `.d.ts` or compilation metadata, compiler CLI plus one generated fixture, native registration and platform names, database path/migration semantics, debugger IDs, and SQLite/build labels. The remaining blocker is this integration choice, not a native API allocation step.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Agreed. The current test observes coordinator lifetime but does not deterministically prove that deferred queue destruction ran off the owned queue. I traced the intended transfer to the resource-release queue; I will add an injected, flushable release queue test that proves the final queue-owning references and SQLite handle are released there before resolving this thread.

src/valdi_modules/src/valdi/client_sql/native/ClientSQLNativeModuleFactory.cpp
bjdodson-openai: Agreed. The public create/migration boundary rejects embedded NUL today, but execRawSQL should still defend its own C-string boundary. I will share the embedded-NUL predicate with this helper and add focused coverage before resolving.

src/valdi_modules/src/valdi/client_sql/BUILD.bazel
bjdodson-openai: Agreed, and the internal ClientSQL context makes this broader than SQLite alone. I will add overridable runtime and validation SQLite labels, preserve the existing internal compiler seam, and make the whole native implementation selectable so an embedder links exactly one factory/runtime. I am leaving this open until the merged internal graph proves that the OSS SQLite and factory are absent when internal providers are selected.

src/valdi_modules/src/valdi/client_sql/BUILD.bazel
bjdodson-openai: Agreed. I will fold the implementation contract coverage into //valdi:test, but keep the implementation-specific native test library selectable so Snap’s internal implementation does not accidentally relink the OSS factory. The runtime integration source also needs to join the existing integration library rather than linking a second copy of JSBridgeTestFixture/RuntimeTestsUtils.

valdi/BUILD.bazel
bjdodson-openai: Agreed; same plan as the companion thread. I will add the ClientSQL runtime test source to the existing //valdi:test integration library and retain the standalone focused target without pulling its duplicate fixture library into the aggregate.

compiler/clientsql/test_clientsql.py
```diff
@@ -0,0 +1,1880 @@
...
+            result = subprocess.run(
+                [
+                    str(CLIENTSQL_TOOLCHAIN_EXECUTABLE),
+                    "-s", str(sql_dir),
+                    "-p", "SharedDb",
+                    "-c", "SharedDb",
+                    "-m", "SharedDb",
+                    "-o", str(output),
+                    "-l", "typescript",
+                ],
```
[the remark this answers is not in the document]
bjdodson-openai: /fp False positive: CLIENTSQL_TEST_TOOLCHAIN_EXECUTABLE is supplied by the Bazel py_test via $(rootpath :clientsql_toolchain), and subprocess.run receives a structured argv list with shell=False (the default). No value is interpreted by a shell.

PR #182, opened 2026-08-28
Cooperative termination mode: drain teardown instead of aborting

clholgat:
Draft for discussion, not for merge. Updated to the cooperative-mode approach (replaces the earlier dispatch-guard version on this branch).

## Problem
Aggressive teardown flips `_isDisposed` from another thread and force-terminates JS execution, aborting in-flight work and racing the JS thread. That's a large blast radius of teardown crashes / UB, and most of it is on host/session-scoped runtimes torn down on logout / scope-exit (not just workers).

## Approach — a configurable mode, default cooperative
One flag on the runtime; cooperative mode drains instead of aborting:
- **Host:** `makeJsThreadDispatchFunction` no longer skips work just because `_isDisposed` is set. In-flight and already-queued JS-thread work runs (the context is still alive, since teardown is serialized after it on the serial queue), then teardown destroys the context and later calls are refused by the liveness check. This effectively restores the pre-aggressive ordering.
- **Worker:** `terminate()`/dtor skip `requestExecutionTermination()` and drain like `self.close()`.

Set the flag false for aggressive termination (stops even frozen JS, at the cost of the teardown races), so the aggressive path stays available and can be rolled out again safely once the root cause is fully addressed.

## Open question
This mitigates by reverting to cooperative behavior. The known trade is the original one: cooperative can't force-stop a genuinely frozen JS worker (it drains, so a busy loop won't tear down). A bounded wait-then-escalate would preserve the frozen-JS guarantee while keeping well-behaved teardown safe. Feedback welcome on whether to build that here or in the deeper fix.

## Tests
`bzl test //valdi:test_integration`: new `cooperativeTeardownDrainsInFlightWorkAggressiveSkipsIt` (all 4 JS engines) asserts cooperative drains in-flight work and aggressive skips it; existing teardown and worker-termination tests stay green.

### DIFF
--- a/valdi/src/valdi/runtime/JavaScript/JavaScriptRuntime.cpp
+++ b/valdi/src/valdi/runtime/JavaScript/JavaScriptRuntime.cpp
@@ -344,6 +344,11 @@ void JavaScriptRuntime::postInit() {
         auto runtimeTweaks = listener->getRuntimeTweaks();
         if (runtimeTweaks != nullptr) {
             _dispatchQueue->setDisableSyncCallsInCallingThread(runtimeTweaks->disableSyncCallsInCallingThread());
+            // Pull the termination mode here too: worker runtimes are created fresh (runtimeCreateWorker)
+            // and only inherit the host's listener, not its pushed tweaks, so Runtime::setRuntimeTweaks
+            // (which targets the host runtime) never reaches them. Reading it at postInit — after the
+            // worker's listener is set — lets a worker honor an aggressive-termination override.
+            setCooperativeTermination(runtimeTweaks->useCooperativeTermination());
         }
     }
 }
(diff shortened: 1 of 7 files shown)


valdi_core/src/valdi_core/cpp/Threading/GCDDispatchQueue.cpp
```diff
@@ -108,8 +109,11 @@ void GCDDispatchQueue::sync(const DispatchFunction& function) {
 
 void GCDDispatchQueue::syncEnter(const DispatchFunction& function) {
     _runningSync = true;
+    // Reset on scope exit so an exception unwinding through function() cannot leave _runningSync
+    // stuck true: a stale true is read by JavaScriptRuntime's dispatch guard to keep a disposed
+    // *async* task from running, so leaking it re-opens the run-against-disposed window.
+    Valdi::Defer resetRunningSync([this]() { _runningSync = false; });
```
scorsin-oai: these couple ones seem to be speculative? we don't really use exceptions

